Issue #56, opened 2023-08-24
Multithreading is not working


mr-eyes:
In `manysearch` and probably the rest of the commands, multi-threading is not working. Even when I hard-code the number of cores, it only works on 2. I noticed that with `htop`

mr-eyes:
Maybe we can define it globally in each function

```rust
rayon::ThreadPoolBuilder::new().num_threads(32).build_global().unwrap();
```

ctb:
The benchmarks suggest otherwise? Unless it's broken in the last few days, which is certainly possible... but seems unlikely!

* https://github.com/sourmash-bio/pyo3_branchwater/issues/7
* https://github.com/sourmash-bio/pyo3_branchwater/issues/10
* https://github.com/sourmash-bio/pyo3_branchwater/issues/12
* https://github.com/sourmash-bio/pyo3_branchwater/issues/22

e.g. in the most recent (#22) we see:
```
        Percent of CPU this job got: 3189%
```

with 32 threads.

So! mystery! What system are you running on?

mr-eyes:
I am running on farm. How did you set the `32` cores while `num_threads` does not seem to be passed to rust as per #53 ? Am I getting something wrong?

ctb:
can you try this:
```
export RAYON_THREADS=32
```

ctb:
ref https://github.com/sourmash-bio/pyo3_branchwater/issues/31

mr-eyes:
It still does not work.
Maybe it behaves like that because I have a small number of target sigs? I have 1K query vs. 15 targets.

ctb:
yes, the 15 targets is likely to be a thing!

here's something that runs on farm with the manysearch code, and definitely does do multithreading... 

background repo: https://github.com/dib-lab/2022-branchwater-benchmarking
integrating with pyo3_branchwater: https://github.com/dib-lab/2022-branchwater-benchmarking/pull/8

ctb:
(how does #57 fix this issue? curious what you figured out ;)

ctb:
FYI large scale benchmarks confirm that something is fishy with `manysearch` and threading -

https://github.com/dib-lab/2022-branchwater-benchmarking/issues/9

mr-eyes:
> (how does #57 fix this issue? curious what you figured out ;)

Here're some details
- I started an `srun` job with `2 cpus` but `squeue` showed allocation of `34 cpus` due to `mem<->cpus` proportionality.
[...]

PR #57, opened 2023-08-25
[MRG] Add `-c` `--cores`

mr-eyes:
default is `0` = full cpu (it considers the case it's running on a limited memory slurm job)

### DIFF
--- a/src/python/pyo3_branchwater/__init__.py
+++ b/src/python/pyo3_branchwater/__init__.py
@@ -3,8 +3,19 @@ import sys
 import argparse
 from sourmash.plugins import CommandLinePlugin
 from sourmash.logging import notify
+import os
 
 from . import pyo3_branchwater
+
+
+def get_max_cores():
+    if 'SLURM_CPUS_ON_NODE' in os.environ:
+        return int(os.environ['SLURM_CPUS_ON_NODE'])
+    elif 'SLURM_JOB_CPUS_PER_NODE' in os.environ:
+        return int(os.environ['SLURM_JOB_CPUS_PER_NODE'].split('x')[0])  # Assumes a simple format; won't handle complex scenarios
+    else:
+        return os.cpu_count()
+
 
 class Branchwater_Manysearch(CommandLinePlugin):
     command = 'manysearch'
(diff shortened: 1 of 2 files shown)


mr-eyes:
Resolves #56

mr-eyes:
Ready for review! @ctb

src/python/pyo3_branchwater/__init__.py
```diff
@@ -3,9 +3,20 @@
 import argparse
 from sourmash.plugins import CommandLinePlugin
 from sourmash.logging import notify
+import os
 
 from . import pyo3_branchwater
 
+
+def get_max_cores():
```
ctb, line 11: maybe we should put a try/except around this to catch failures, and just print traceback, report "things failed, returning OS CPU count" on failure?

src/python/pyo3_branchwater/__init__.py
```diff
@@ -24,11 +35,23 @@ def __init__(self, p):
...
                        help='scaled factor at which to do comparisons')
+        p.add_argument('-c', '--cores', default=0, type=int,
+                       help='number of cores to use (default is all available)')
 
     def main(self, args):
         notify(f"ksize: {args.ksize} / scaled: {args.scaled} / threshold: {args.threshold}")
-        num_threads = pyo3_branchwater.get_num_threads()
+
+
+        avail_threads = get_max_cores()
```
ctb, line 45: this code could be combined into a single function since it's the same everywhere, I think.

ctb:
https://github.com/mr-eyes/pyo3_branchwater/pull/1 cleans up some spaces ;).

Curious about removing num threads - I guess we don't need it any more? But it seems nice to have.

src/python/pyo3_branchwater/__init__.py
ctb: also, what about `RAYON_THREADS`?

mr-eyes: I couldn't find information about `RAYON_THREADS`. Could you please explain more?

mr-eyes: Oh, you meant this.
https://github.com/sourmash-bio/pyo3_branchwater/issues/56#issuecomment-1692370679
This is useless now since it's fully controlled through the code.

ctb:
thank you!

PR #64, opened 2023-08-26
EXP: break into modules

### LINKED ISSUE #62: consider splitting `lib.rs` into multiple files...
it's getting a bit unwieldy!

ctb:
Experimental.

This is a first cut at breaking things up like so:
```
src/lib.rs - pyo3 code
src/manysearch.rs - the 'manysearch' function
src/utils.rs - all of the functions used in `manysearch.rs`, etc.
```

If this is the structure we go with, there would be one file for each top-level command, e.g. `src/fastgather.rs`.

Fixes #62.


ctb:
done in #101.

PR #76, opened 2023-08-30
[MRG] add max_containment column

### LINKED ISSUE #75: Adding maximum containment column
I will be working on adding a maximum containment column.

### LINKED ISSUE #77: add jaccard back into manysearch
In #52 / fixing #25, I added Jaccard calculation/output into manysearch.

this ended up costing a lot of compute time, because my approach was to merge the minhashes and then calculate their length; see #71.

so, in #72 / fixing https://github.com/sourmash-bio/pyo3_branchwater/issues/71, I removed Jaccard.

But I think we can put it back in:

Upon reflection, we don't need to merge the two minhashes - this is something that (I think) is needed for handling num MinHash and is not needed for FracMinHash. For FracMinHash, we can use `num_common_hashes / (size_A + size_B - num_common_hashes)` which should be cheap.

### LINKED ISSUE #83: print out nicer gather stats (remaining, etc)

### LINKED ISSUE #84: print out branchwater version when running

mr-eyes:
resolves #75 

also: because of #85,

Fixes #83 
Fixes #77 
Fixes #84


ctb (changes requested):
please add a test for the specific numbers output, over in `src/python/tests/test_search.py`. thanks!

ctb:
tests added in https://github.com/sourmash-bio/pyo3_branchwater/pull/85

PR #86, opened 2023-09-01
[MRG] abstract writer thread into reusable function

bluegenes:
Since we reuse the multithreading code in each function (and I wanted to understand it better), this PR abstracts the writer thread into a reusable function and uses it in both `manysearch` and `mastiff_manysearch`.

Ideally, we could reuse this for all functions (gather, etc) -- but, I wanted a review to see if this went in the right direction.

Since this change means that we need to pass a consistent object to the writer, this PR also creates a `SearchResult` struct (and associated methods) that is created during each 'search' loop and passed to the writer thread. 

For gather (a later PR if we want to go this route), Luiz already has a GatherResult in the works (https://github.com/sourmash-bio/sourmash/blob/ba581ea8c6f046b7e19da274155f326d2bdbc8e7/src/core/src/index/revindex.rs#L572), so it's possible we could import from there. But defining a simple one here while things are still in flux may be the easier option, since I want it to have these formatting options for easy printing...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,6 +33,14 @@ use sourmash::index::revindex::{RevIndex};
 use sourmash::prelude::MinHashOps;
 use sourmash::prelude::FracMinHashOps;
 
+
+// initialize logging
+fn initialize_logger() {
+    env_logger::Builder::new()
+        .filter_level(log::LevelFilter::Info) // Set the desired log level
+        .init();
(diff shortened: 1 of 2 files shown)


bluegenes:
@ctb ready for review (and any associated discussion :)

ctb (approved):
Looks great!

PR #88, opened 2023-09-02
MRG: enable multithreaded sketching to zip file (manysketch)

bluegenes:
Includes #96.

Enable manysketch to zipfile (that can then be used to make rocksdb index).

This PR adds:
- write sourmash manifest for zipfile
- handle identical md5sums (add `_i` to filenames and make sure it's reflected in the internal_location)
- use param string(s) to allow the full range of sketch values.
- take in sketch fromfile input
- properly name sigs from sketch fromfile
- adjust protein ksize*3 for proper prot sketching

Notes:
- Adding a manifest required a bit more tricky thread handling (and use of Arc), since we need to collect all the manifest info from the signatures while looping, and then write the manifest at the end, just before closing the zip and joining the writer thread.
- Only builds dna --> dna, protein--> protein sigs (not translated). Translated could be added in the future.
- does not allow param string input from fromfile_csv

## Usage

To run `manysketch`, you need to build a csv file with columns `name,genome_filename,protein_filename` (`fa.csv`, below). You can then run:
```
sourmash scripts manysketch fa.csv -o fa.zip
```
The output will be written to `fa.zip`

You can check if all signatures were written properly with
```
sourmash sig summarize fa.zip
```

To modify sketching parameters, use `--param-str` or `-p` and provide valid param string(s)
```
sourmash scripts manysketch fa.csv -o fa.zip -p k=21,k=31,k=51,scaled=1000,abund -p protein,k=10,scaled=200
```


bluegenes:
@ctb ready for review

src/lib.rs
```diff
@@ -1503,9 +1681,25 @@ fn do_multisearch(querylist_path: String,
         Err(e) => {
             eprintln!("Error: {e}");
             Ok(1)
+            }
         }
     }
```
bluegenes, line 1686:
```suggestion
        }
    }
}
```

ctb:
nifty! will dig in later, but one request and one suggestion -

* request: could you update the docs with some minimal execution info?
* suggestion: could you give some benchmarks?

well, and a third thought - might be good to start an issue with upgrade suggestions, like "enable protein sketching" or whatever comes to mind.

ctb (approved):

PR #96, opened 2023-09-05
MRG: add zipfile manifest, file and sketch checking, etc

bluegenes:
note: PR into #88.

This PR adds:
- write sourmash manifest for zipfile
- handle identical md5sums (add `_i` to filenames and make sure it's reflected in the internal_location)
- use param string(s) to allow the full range of sketch values.
- take in sketch fromfile input
- properly name sigs from sketch fromfile
- adjust protein ksize*3 for proper prot sketching

Notes:
- Adding a manifest required a bit more tricky thread handling (and use of Arc), since we need to collect all the manifest info from the signatures while looping, and then write the manifest at the end, just before closing the zip and joining the writer thread.
- Only builds dna --> dna, protein--> protein sigs (not translated). Translated could be added in the future.
- does not allow param string input from fromfile_csv

## Usage

To run `manysketch`, you need to build a csv file with columns `name,genome_filename,protein_filename` (`fa.csv`, below). You can then run:
```
sourmash scripts manysketch fa.csv -o fa.zip
```
The output will be written to `fa.zip`

You can check if all signatures were written properly with
```
sourmash sig summarize fa.zip
```

To modify sketching parameters, use `--param-str` or `-p` and provide valid param string(s)
```
sourmash scripts manysketch fa.csv -o fa.zip -p k=21,k=31,k=51,scaled=1000,abund -p protein,k=10,scaled=200
```


bluegenes:
ok @ctb ready for review

ctb:
(did you want to keep this separate from #88 for a reason?)

bluegenes:
@ctb - nope, can merge and review all at once if you prefer!

PR #99, opened 2023-09-07
MRG: enable zipfile loading for manysearch

bluegenes:
Follow up to #105: 
- Enables sourmash zipfiles to be used as input to `manysearch` and `mastiff_manysearch`.

to do:
- [x] use file loading functions from #105 
- [x] modify tests as needed for unified reporting logic
- [x] add tests for loading from zipfiles

Notes:
When we build pathlists from zipfiles, internal sketches are extracted to individual files. To prevent excessive and unhelpful warnings, we only warn about individual input files without compatible sketches if the input was not a zipfile.


ctb:
hi @bluegenes I did some major code moving in https://github.com/sourmash-bio/pyo3_branchwater/pull/101, and hope to merge that ASAP. Apologies but `lib.rs` was getting a bit out of control 😆

bluegenes:
No worries, it was definitely getting crazy.

This PR should be fairly easy to adapt to your new structure.

src/python/tests/test_search.py
```diff
@@ -447,23 +504,33 @@ def test_load_only_one_bug_as_query(runtmp, capfd, indexed):
...
     print(captured.err)
+    print(runtmp.last_result.out)
 
     assert not 'WARNING: skipped 1 paths - no compatible signatures.' in captured.err
-    assert not 'WARNING: no compatible sketches in path ' in captured.err
+    # NTP:
+    # This fails with zip input, I think bc each sig becomes an individual file when zipped
+    # todo - do we want to fix this for less verbose reporting when passing a zip file?
+    if not zip_query:
+        assert not 'WARNING: no compatible sketches in path ' in captured.err
```
bluegenes, line 528: When we read pathlists from zipfiles, independent signatures stored in more than one file become individual files in the zip. So here, where we are loading each individual path from the pathlist, we may get a lot of 'WARNING's. Do we want to add a zip boolean to prevent excessive reporting?

bluegenes:
@ctb ready for review!

ctb (approved):

PR #105, opened 2023-09-08
MRG: add utility functions for zip reading; use in index, multisearch

bluegenes:
Extracted from #99, because all the changes were getting unruly.

This should be easier to review - contains the functions & modifications to use them in `index` and `multisearch`, with tests.

New in `utils.py`:
- load_sigpaths_from_zip_or_pathlist
- load_sketches_from_zip_or_pathlist
- report_on_sketch_loading (unifies signature loading reporting)

Renamed:
- `read_signatures_from_zip` --> `load_sketches_from_zip`
- `load_sketch_fromfile` --> `load_fasta_fromfile`

Note: In the case of reading a pathlist from a zip file, this means we extract the sig files and store them in a temp directory which will be deleted when it goes out of scope. This adds a little overhead, but I think it could worth it for the simplicity of being able to specify a zip database. Open to removing, though.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -221,6 +221,31 @@ pub fn load_sketchlist_filenames<P: AsRef<Path>>(sketchlist_filename: &P) ->
 }
 
 
+/// Loads signature file paths from a ZIP archive.
+///
+/// This function extracts the contents of a ZIP archive containing
+/// signature files (with extensions ".sig" or ".sig.gz") to a temporary directory.
+/// It returns the paths of these extracted signature files.
+///
+/// # Arguments
+///
+/// * `zip_path` - The path to the ZIP archive.
+///
+/// # Returns
+///
+/// Returns a tuple containing:
+/// * A vector of `PathBuf` representing the paths to the extracted signature files.
+/// * The `TempDir` representing the temporary directory where the files were extracted.
+///   Since tempfile::TempDir creates a temporary directory that is automatically
+///   deleted once the TempDir value goes out of scope, we return it here to move it
+///   to the main function scope.
+///
+/// # Errors
+///
+/// Returns an error if:
+/// * Unable to create a temporary directory.
+/// * Unable to open or read the ZIP archive.
+/// * Any other IO or file related error.
 pub fn load_sigpaths_from_zip<P: AsRef<Path>>(
     zip_path: P,
 ) -> Result<(Vec<PathBuf>, tempfile::TempDir)> {
@@ -246,6 +271,7 @@ pub fn load_sigpaths_from_zip<P: AsRef<Path>>(
     println!("loaded paths for {} signature files from zipfile {}", signature_paths.len(), zip_path.as_ref().display());
     Ok((signature_paths, temp_dir))
 }
+
 
 pub fn load_fasta_fromfile<P: AsRef<Path>>(sketchlist_filename: &P) -> Result<Vec<(String, PathBuf, String)>> {
     let mut rdr = csv::Reader::from_path(sketchlist_filename)?;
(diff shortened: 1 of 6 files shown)


bluegenes:
@ctb this is ready for review

src/utils.rs
```diff
@@ -375,6 +401,127 @@ pub fn load_sketches_above_threshold(
...
+
+        if !file_name.ends_with(".sig") && !file_name.ends_with(".sig.gz") {
+            continue;
+        }
+        if let Ok(sigs) = Signature::from_reader(&mut file) {
+            if let Some(sm) = prepare_query(&sigs, template, &zip_path.as_ref().display().to_string()) {
+                sketchlist.push(sm);
+            } else {
+                // track number of paths that have no matching sigs
+                skipped_paths.fetch_add(1, atomic::Ordering::SeqCst);
```
ctb, line 428: do we need atomic types? This loop doesn't look like it's parallelized, right?

ctb:
I like overall, but want to dig in to code a bit.

src/utils.rs
bluegenes: Nope, not parallelized. 

There is a parallel zip library we could use if we want, but probably not needed except when reading from genbank-scale zips.

bluegenes: removed atomic in [1ea65be](https://github.com/sourmash-bio/pyo3_branchwater/pull/105/commits/1ea65be1e3935679e5d389a236e5271c4835c3d1)

src/utils.rs
```diff
@@ -221,7 +220,34 @@ pub fn load_sketchlist_filenames<P: AsRef<Path>>(sketchlist_filename: &P) ->
     Ok(sketchlist_filenames)
 }
 
-pub fn load_sketch_fromfile<P: AsRef<Path>>(sketchlist_filename: &P) -> Result<Vec<(String, PathBuf, String)>> {
+
+pub fn load_sigpaths_from_zip<P: AsRef<Path>>(
```
ctb, line 224: note to self: this function name is confusing. It reads like we are loading a sigpath file from a zip. Can't come up with a better name tho 😄 

I think the best solution would be to comment it some more. Will embark.

src/utils.rs
```diff
@@ -221,7 +220,34 @@ pub fn load_sketchlist_filenames<P: AsRef<Path>>(sketchlist_filename: &P) ->
...
+
+    for i in 0..zip_archive.len() {
+        let mut file = zip_archive.by_index(i)?;
+        let mut sig = Vec::new();
+        file.read_to_end(&mut sig)?;
+
+        let file_name = Path::new(file.name()).file_name().unwrap().to_str().unwrap();
+        if file_name.ends_with(".sig") || file_name.ends_with(".sig.gz") {
+            let mut new_file = File::create(temp_dir.path().join(file_name))?;
+            new_file.write_all(&sig)?;
```
ctb, line 240: THIS writes a new copy of the sig to a temp dir. (How is that temp dir chosen?)

src/utils.rs
```diff
@@ -375,6 +401,127 @@ pub fn load_sketches_above_threshold(
...
+    // loop through, loading signatures
+    for i in 0..zip_archive.len() {
+        let mut file = zip_archive.by_index(i)?;
+        let file_name = Path::new(file.name()).file_name().unwrap().to_str().unwrap().to_owned();
+
+        if !file_name.ends_with(".sig") && !file_name.ends_with(".sig.gz") {
+            continue;
+        }
+        if let Ok(sigs) = Signature::from_reader(&mut file) {
+            if let Some(sm) = prepare_query(&sigs, template, &zip_path.as_ref().display().to_string()) {
```
ctb, line 424: THIS loads directly from zip into memory.

src/utils.rs
bluegenes: I'll write some docstrings

src/utils.rs
bluegenes: https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html
>The [TempDir](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html) type creates a directory on the file system that is deleted once it goes out of scope. At construction, the TempDir creates a new directory with a randomly generated name.

> The default constructor, [TempDir::new()](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.new), creates directories in the location returned by [std::env::temp_dir()](https://doc.rust-lang.org/std/env/fn.temp_dir.html), but TempDir can be configured to manage a temporary directory in any location by constructing with a [Builder](https://docs.rs/tempfile/latest/tempfile/struct.Builder.html).


potentially important (and not yet handled):
> Note that if the program exits before the TempDir destructor is run, such as via [std::process::exit()](http://doc.rust-lang.org/std/process/fn.exit.html), by segfaulting, or by receiving a signal like SIGINT, then the temporary directory will not be deleted.

src/utils.rs
bluegenes: added in [6807233](https://github.com/sourmash-bio/pyo3_branchwater/pull/105/commits/68072335bb30c6403a400138b9c05918777939c5)

ctb: nice!

src/utils.rs
ctb: I think we're fine with this!

ctb (approved):
Nice work! good code simplification too 👍

PR #106, opened 2023-09-08
MRG: add zip reading to `fastmultigather`

bluegenes:
Follow up to #105: 
- Enables sourmash zipfiles to be used as input to `fastmultigather`.

to do:
- [x] use file loading functions from #105 
- [x] modify tests as needed for unified reporting logic
- [x] add tests for loading from zipfiles

Notes:
- When we build pathlists from zipfiles, internal sketches are extracted to individual files. To prevent excessive and unhelpful warnings, we only warn about individual input files without compatible sketches if the input was not a zipfile. 
- This also means tests need to be different - sig files will be named by their md5 instead of the zipfile filename


bluegenes:
@ctb ready for review

ctb (approved):

PR #121, opened 2023-09-20
MRG: clearer message for manysketch

bluegenes:
For a small csv file (especially with large files), the prior printout was misleading. After opening one file per thread, the message claims x% done. In reality, we've just opened the files to start building sketches.

### DIFF
--- a/src/manysketch.rs
+++ b/src/manysketch.rs
@@ -187,7 +187,10 @@ pub fn manysketch<P: AsRef<Path> + Sync>(
             // progress report at threshold
             if i != 0 && i % reporting_threshold == 0 {
                 let percent_processed = ((i as f64 / n_fastas as f64) * 100.0).round();
-                eprintln!("Starting {}th fasta file ({}% of total)", i, percent_processed);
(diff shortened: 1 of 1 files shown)


src/manysketch.rs
```diff
@@ -187,7 +187,10 @@ pub fn manysketch<P: AsRef<Path> + Sync>(
             // progress report at threshold
             if i != 0 && i % reporting_threshold == 0 {
                 let percent_processed = ((i as f64 / n_fastas as f64) * 100.0).round();
-                eprintln!("Processed {} fasta files ({}% done)", i, percent_processed);
+                eprintln!(
+                    "Starting {}th fasta file ({}% of total)",
```
ctb, line 191: I'm OK with this but it is a bit weird to see 1th, 2th, 3th :)

Could do {}/{} ({}% of total)

ctb (approved):

src/manysketch.rs
bluegenes: Oh, that's definitely better- will add

PR #124, opened 2023-09-21
MRG: fix clippy stuff; bump to v0.8.1

ctb:
This PR cleans up several clippy suggestions, and also bumps the version to v0.8.1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1251,7 +1251,7 @@ dependencies = [
 
 [[package]]
 name = "pyo3-branchwater"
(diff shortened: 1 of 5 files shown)


ctb:
@bluegenes can you take a quick look? thanks :)

ctb:
fun factoid: `_temp_dir` isn't used explicitly but its transfer of ownership to the top level function scope is important, because when it goes out of scope it deletes the temp directory! This causes tests to fail on the CI (Linux), but not Mac.

PR #134, opened 2023-09-27
MRG: use updated rocksdb-integrated sourmash core

bluegenes:
This is entirely a developer-focused update to level-set the codebase.

Since the addition of rocksdb-indexed gather to this plugin, we have been relying on a prior version of the sourmash core, without the final updates/ refactoring that went into https://github.com/sourmash-bio/sourmash/pull/2230. Sourmash core 0.12.0, our latest release, did not have automatic downsampling while selecting signatures, which broke some of our functionality. After https://github.com/sourmash-bio/sourmash/pull/2931, we can now use the updated core code here.

Notes:
- currently uses sourmash core within sourmash `latest`, pinned to current commit. After releasing a new core, we should update to that.
- Where code was working and did not rely on changed core functionality, I didn't change anything. Instead, I focused only on code that needed to be changed --  primarily any code using a rocksdb index.
- Within there, main changes:
  - use `selection` and `Signature::Select` rather than template-based selection
  - in `index`, use new `Collection` to load zipfile `Collection::from_zipfile(siglist)?`
  - adapt code for calling `gather`, to reflect adjusted names/ structs from sourmash core as necessary

We can later use the updated `core` code to improve /standardize more functionality here, see https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/196.

### Behavior changes:
- `index`: If there are no signatures to index, now produces empty index file instead of erroring out. 
> Is this desired? This mirrors what happens within `core`, but we could add a check to see if the index is empty and throw an error from here.
> -  to do this, we would need an easy check to see if the `rocksdb` index is empty. @luizirber -- does `db.check()` do that?

PR #136, opened 2023-10-05
MRG: read `sig.gz_{n}` sigs from zipfiles

bluegenes:
as noted in https://github.com/sourmash-bio/sourmash/issues/2749#issuecomment-1715762525, sourmash zipfiles store repeated md5sums as `.sig.gz_{n}` instead of `_{n}.sig.gz`. This PR allows us to read those files as well, rather than skipping over them.

### DIFF
--- a/src/mastiff_manygather.rs
+++ b/src/mastiff_manygather.rs
@@ -94,10 +94,8 @@ pub fn mastiff_manygather<P: AsRef<Path>>(
                         let threshold = threshold_bp / query.minhash.scaled() as usize;
 
                         // mastiff gather code
-                        println!("Building counter");
                         let (counter, query_colors, hash_to_color) =
                             db.prepare_gather_counters(&query.minhash);
-                        println!("Counter built");
 
                         let matches = db.gather(
                             counter,
--- a/src/python/tests/test_index.py
+++ b/src/python/tests/test_index.py
@@ -187,6 +187,35 @@ def test_index_zipfile(runtmp, capfd):
     assert 'Found 3 filepaths' in captured.err
 
 
+def test_index_zipfile_repeated_md5sums(runtmp, capfd):
+    # test that we're reading all files, including repeated md5sums
+    siglist = runtmp.output('db-sigs.txt')
+
+    sig2 = get_test_data('2.fa.sig.gz')
+    sig2a = runtmp.output('sig2a.sig.gz')
+    sig2b = runtmp.output('sig2b.sig.gz')
+    runtmp.sourmash('sig', 'rename', sig2, 'name2', '-o', sig2a)
+    runtmp.sourmash('sig', 'rename', sig2, 'name3', '-o', sig2b)
+
+    make_file_list(siglist, [sig2, sig2a, sig2b])
+
+    zipf = runtmp.output('sigs.zip')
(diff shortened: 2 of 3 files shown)


ctb (approved):
I mean, ok, but ... I worry we're now introducing different behavior b/t pyo3_branchwater and sourmash :)

Two questions:
* how does manysketch handle this situation? Are multiple sketches created and saved?
* does pyo3_branchwater currently pay any attention to manifests when reading?

Anyway, I think it's fine to go ahead and merge, but let's also create an issue to re-examine this in the future :)

bluegenes:
> * how does manysketch handle this situation? Are multiple sketches created and saved?

many sketch writes multiple signatures, using `format!("signatures/{}_{}.sig.gz", md5sum_str, count)` for duplicates. I had thought this was replicating what we do in sourmash, didn't realize we were appending the count _after_ the `gz`. 

relevant `utils.rs` code for writing sigs from `manysketch`
```
 for (sig, param) in sigs.iter().zip(params.iter()) {
                        let md5sum_str = sig.md5sum();
                        let count = md5sum_occurrences.entry(md5sum_str.clone()).or_insert(0);
                        *count += 1;
                        let sig_filename = if *count > 1 {
                            format!("signatures/{}_{}.sig.gz", md5sum_str, count)
                        } else {
                            format!("signatures/{}.sig.gz", md5sum_str)
                        };
                        write_signature(sig, &mut zip, options, &sig_filename);
                        manifest_rows.push(make_manifest_row(
                            sig,
                            &filename,
                            &sig_filename,
                            param.scaled,
                            param.num,
                            param.track_abundance,
                            param.is_dna,
                            param.is_protein,
                        ));
```

> * does pyo3_branchwater currently pay any attention to manifests when reading?

no, it currently looks through the sig files themselves. I expect to switch to using manifests as part of #134 or follow ups.


Hmm, the optimal solution might be to use the manifest to load the same file for duplicates (not storing the duplicates at all). What are the challenges associated with that solution?

ctb:
> Hmm, the optimal solution might be to use the manifest to load the same file for duplicates (not storing the duplicates at all). What are the challenges associated with that solution?

ooh! I like it ;). Not ready to commit to it on sourmash yet, but hot take is it's a leading contender!

bluegenes:
Note: some of our utilities may only keep `*sig.gz` files, leaving duplicates behind

`sig extract` does not keep duplicates:
```
sourmash sig extract -k 21 mm.dna.zip -o mm.dna-k21.zip

== This is sourmash version 4.8.4. ==
== Please cite Brown and Irber (2016), doi:10.21105/joss.00027. ==

loaded 428 total that matched ksize & molecule type
extracted 328 signatures from 1 file(s)
```

`sourmash sig grep`, _does_ keep duplicate sigs
```
sourmash sig grep mammarenavirus spillover.dna.zip -k 21 -o mm.dna.zip

== This is sourmash version 4.8.4. ==
== Please cite Brown and Irber (2016), doi:10.21105/joss.00027. ==

saving matching signatures to 'mammarenavirus2.dna.zip'
loaded 33566 total that matched ksize & molecule type
extracted 428 signatures from 1 file(s)
```

ctb:
(just to be clear, I think you can/should merge this :)

PR #157, opened 2023-11-29
MRG: fix error for large manysketch zips (> 4G)

### LINKED ISSUE #147: Crash with sourmash manysketch with very large (2 billion row+) datasets
Hello, 
I was using manysketch on a large dataset, and noticed that on the manifest writing step, it crashed saying that zipwriter was already closed. Doing some digging, I think the issue is that the options for zip::write::FileOptions uses the default large file setting (false) here https://github.com/sourmash-bio/pyo3_branchwater/blob/3f1994365c40d57c1fc3d6040b4bd5d2e8c8f5e2/src/utils.rs#L1019 
which causes an exception to get thrown if the zip file ends up being over 4gb (https://docs.rs/zip/latest/zip/write/struct.FileOptions.html#method.large_file). Setting options to `zip::write::FileOptions::default().compression_method(zip::CompressionMethod::Stored).large_file(true);` seems to have fixed the issue

bluegenes:
Fixes #147

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1016,7 +1016,7 @@ pub fn sigwriter<P: AsRef<Path> + Send + 'static>(
         let file_writer = open_output_file(&output);
 
         let options =
-            zip::write::FileOptions::default().compression_method(zip::CompressionMethod::Stored);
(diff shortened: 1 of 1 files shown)


bluegenes:
ready for review @ctb

ctb (approved):

PR #180, opened 2024-01-07
EXP: support output of details of gather

ctb:
This adds CSV output containing the number of matches and the number of hashes remaining at each stage of fastgather.


ctb:
I think the main change to make here is to adjust the rank etc. so that it's the same as in the gather output. It might already be, just need to double check.

ctb:
Closing this now that v0.9.5 has all the gather things!!

PR #184, opened 2024-01-08
MRG: enable `singleton` sketching, facilitate `reads`-based sketching

bluegenes:
Here I modify `manysketch` to allow:
- sketching multiple input files into the same sketch
- sketching singletons!

New Functions
- `detect_csv_type` - figure out which type of acceptable input csv we have
- `process_assembly_csv` - process standard `fromfile` csv (headers: `name,genome_filename,protein_filename`). This was originally part of `load_fasta_fromfile`.
- `process_reads_csv` - process new `reads` csv (headers: `name,read1,read2`)

This refactoring also makes it much simpler for us to add new types of CSVs that can be read.

Notes:
- parallelization is still done by each set of inputs that should yield signatures. If singleton, we're sketching all records for a given file in series, not all sketches in parallel.
- perhaps a confusing thing: when sketching with `--singleton`, the name entry in each row doesn't end up getting used, since we name from the records themselves.

## 1. Multiple input files for a single sketch

We can now sketch multiple filenames into the same sketch. **This is currently only possible if you input a 'reads' csv, with format `name,read1,read2`, which is designed to add support for metagenomes with PE reads.** Note that the `read2` column name must be present, but can be empty. Since this is designed for metagenome reads, we assume input files are DNA.

The original `sketch fromfile` input format, `name,genome_filename,protein_filename` remains functional.

Future work could introduce additional ways to pass more than one filename per sketch.

## 2. Singleton sketching

With either input filetype, we can now pass `--singleton` and get one sketch per record, named from the record name.

Prefix-based sketching punted to #243

### DIFF
--- a/src/manysketch.rs
+++ b/src/manysketch.rs
@@ -168,6 +168,22 @@ pub fn manysketch(
         .par_iter()
         .filter_map(|(name, filenames, moltype)| {
             let mut allsigs = Vec::new();
+            // build sig templates for these sketches from params, check if there are sigs to build
+            let sig_templates = build_siginfo(&params_vec, moltype);
+            // if no sigs to build, skip this iteration
+            if sig_templates.is_empty() {
+                skipped_paths.fetch_add(filenames.len(), atomic::Ordering::SeqCst);
+                processed_fastas.fetch_add(1, atomic::Ordering::SeqCst);
+                return None;
+            }
+
+            let mut sigs = sig_templates.clone();
+            // have name / filename been set for each sig yet?
+            let mut set_name = false;
+            // if merging multiple files, sourmash sets filename as last filename
+            let last_filename = filenames.last().unwrap();
+
+            // to do: consider changing reporting to per-sig, no matter how many fastas? but singleton...
             for filename in filenames {
                 // increment processed_fastas counter; make 1-based for % reporting
                 let i = processed_fastas.fetch_add(1, atomic::Ordering::SeqCst);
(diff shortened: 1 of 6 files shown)


src/manysketch.rs
```diff
@@ -2,13 +2,153 @@
...
+    {
+        CSVType::Assembly
+    } else if headers.len() == 3
+        && headers.get(0).unwrap() == "name"
+        && headers.get(1).unwrap() == "read1"
+        && headers.get(2).unwrap() == "read2"
+    {
+        CSVType::Reads
+    } else {
+        CSVType::Unknown
```
bluegenes, line 34: add: `CSVType::prefix`, where we read in fasta files as described by a prefix that may/may not include regex?

(as in https://github.com/dib-lab/sourmash-slainte/issues/11)

bluegenes:
@ctb ready for review.

src/python/sourmash_plugin_branchwater/__init__.py
```diff
@@ -336,6 +336,8 @@ def __init__(self, p):
                           help='parameter string for sketching (default: k=31,scaled=1000)')
         p.add_argument('-c', '--cores', default=0, type=int,
                        help='number of cores to use (default is all available)')
+        p.add_argument('-s', '--singleton', action="store_true",
+                       help='build one sketch per fasta record')
```
ctb, line 340:
```suggestion
                       help='build one sketch per FASTA record, i.e. multiple sketches per FASTA file')
```

ctb (approved):
Looks great, works great!

Some minimal documentation and an update to the help docstring would be lovely :). But punt to an issue if you prefer to add those later.

PR #211, opened 2024-02-13
MRG: enable loading manifest files

bluegenes:
Refactor sequential loading attempts in `load_collection` for individual functions that can propagate errors. Add `collection_from_manifest` to allow loading a manifest file directly.

Loading a manifest is a much better alternative to loading a pathlist, since it allows lazy loading -- i.e. we don't need to load all signatures immediately to obtain parameters for selection.


ctb:
(these are standalone manifests as output by `sourmash sig manifest`, right? or something else?)

ctb:
ahh! I see it in the docs :)

ctb (approved):

ctb:
nice!

Issue #214, opened 2024-02-15
gather/fastgather benchmarks - feb 15, 2024


ctb:
on SRR606249 at k=31, scaled=1000

## Summary

[hackmd version](https://hackmd.io/d4NP2i1mRjC8ZAaxQX2ofw?view)

<img width="912" alt="Screenshot 2024-02-19 at 4 54 39 PM" src="https://github.com/sourmash-bio/sourmash_plugin_branchwater/assets/51016/3f95737c-9663-4585-aec5-6b6c2c7f1ef9">

## fastgather with zip file

v0.9.0-pre, as of #197/#211 merge (using sourmash core v0.12.1)

* 2m 5s
* 14 GB RAM

```
        Command being timed: "sourmash scripts fastgather SRR606249.trim.k31.sig.zip /group/ctbrowngrp/sourmash-db/gtdb-rs214/gtdb-rs214-k31.zip -o SRR606249.x.gtdb-rs214.fastgather.csv"
        User time (seconds): 6023.69
        System time (seconds): 31.83
        Percent of CPU this job got: 4837%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:05.17
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 14118520
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 974
        Minor (reclaiming a frame) page faults: 2421007
        Voluntary context switches: 45047
        Involuntary context switches: 151667
        Swaps: 0
        File system inputs: 199312
        File system outputs: 56
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

ctb:
## sourmash gather

as of v4.8.7

45m 38s
14.5 GB of RAM

```
        Command being timed: "sourmash gather -k 31 SRR606249.trim.k31.sig.gz /group/ctbrowngrp/sourmash-db/gtdb-rs214/gtdb-rs214-k31.zip -o SRR606249.x.gtdb-rs214.csv"
        User time (seconds): 2644.60
        System time (seconds): 45.33
        Percent of CPU this job got: 98%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 45:38.43
        Average shared text size (kbytes): 0
[...]

PR #221, opened 2024-02-17
MRG: parallelize `collection_from_pathlist`

### LINKED ISSUE #220: `collection_from_pathlist` is much slower than the other methods
In the new `load_collection` framework, it looks like `collection_from_pathlist` introduced in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/211 does not read sketches in parallel, leading to extremely long load times for pathlists.

Here is the relevant benchmark result from https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/214:

Screenshot 2024-02-17 at 3 08 21 PM

We would still expect it to be slower, but maybe by a factor of two - not 18x! :)

ctb:
This PR updates `collection_from_pathlist` to first load the list of paths, and then load them all in parallel; previously, in #211, the sketches were all loaded in serial, resulting in an 18x fold slowdown vs other methods.

Benchmark proof: https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/214#issuecomment-1950646175

It also fixes a test to write manifests to the temp directory rather than the `test-data` directory.

Fixes #220

### DIFF
--- a/src/python/tests/test_gather.py
+++ b/src/python/tests/test_gather.py
@@ -669,8 +669,8 @@ def test_simple_with_manifest_loading(runtmp):
     sig63 = get_test_data('63.fa.sig.gz')
 
     make_file_list(against_list, [sig2, sig47, sig63])
-    query_manifest = get_test_data("query-manifest.csv")
-    against_manifest = get_test_data("against-manifest.csv")
+    query_manifest = runtmp.output("query-manifest.csv")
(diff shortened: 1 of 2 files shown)


ctb:
Ready for review and merge @bluegenes !

ctb:
or @mr-eyes ;)

mr-eyes (approved):
Thanks for the mention. It looks good to me!

Issue #219, opened 2024-02-17
Adapt the kSpider's algorithm in pairwise comparisons


mr-eyes:
I plan to implement the kSpider's algorithm in Rust, utilizing the inverted index of RocksDB. I wanted to ask if that should be another sub-command in the `branchwater pairwise` or if it should be a separate command. Or maybe a different repo (i.e., the original kSpider project).

mr-eyes:
The kSpider algorithm is particularly good in comparing large **sparse** hash-sets.

ctb:
A few hot takes:
* I think in this repo (branchwater plugin) is fine, as long as it is clearly documented and tested 😁 
* if it takes unique/restricted/limited database types, e.g. ONLY a rocksdb index, I would suggest a new command
* if it behaves EXACTLY like pairwise and can autodetect a rocksdb, then leave it as pairwise
* note that the rocksdb database format has been evolving a bit, e.g. b/t v0.8.6 and 0.9.0 of this plugin, it has changed. That may be because of the update to sourmash core.

mr-eyes:
Hi @bluegenes , I would absolutely love to contribute to this as needed. I want to clarify that #234 is not what I mean by the kSpider's algorithm. The kSpider algorithm utilizes the index to generate a sparse pairwise matrix rather than the brute-force comparisons done in #181. This is expected to work well when comparing highly diverse/sparse data.

bluegenes:
Whoops - completely misread that somehow. @ctb's comments make a lot more sense now 😂. Sparse pairwise would be great too!

In any case, would love to have your input on the cluster PR :). I 'll tag your for review when I think it's ready.

PR #234, opened 2024-02-22
MRG: Add graph-based clustering

bluegenes:
This PR adds a new command, `cluster`, that can be used to cluster the output from `pairwise` and `multisearch`.

`cluster`uses `rustworkx-core` (which internally uses `petgraph`) to build a graph, adding edges between nodes when the similarity exceeds the user-defined threshold. It can work on any of the similarity columns output by `pairwise` or `multisearch`, and will add all nodes to the graph to preserve singleton 'clusters' in the output.

`cluster` outputs two files: 
1. cluster identities file: `Component_X, name1;name2;name3...`
2. cluster size histogram `cluster_size, count`

context for some things I tried:
- try using petgraph directly and removing rustworkx dependency
> nope,`rustworkx-core` adds `connected_components` that returns the connected components, rather than just the number of connected components. Could reimplement if `rustworkx-core` brings in a lot of deps
- try using 'extend_with_edges' instead of add_edge logic.
> nope, only in `petgraph`

**Punted Issues:**
- develop clustering visualizations (ref @mr-eyes kSpider/dbretina work). Optionally output dot file of graph? (https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/248)
- enable updating clusters, rather than always regenerating from scratch (https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/249)
- benchmark `cluster` (https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/247)
>  `pairwise` files can be millions of lines long. Would it be faster to parallel read them, store them in an `edges` vector, and then add nodes/edges sequentially? Note that we would probably need to either 1. store all edges, including those that do not pass threshold) or 2. After building the graph from edges, add nodes from `names_to_node` that are not already in the graph to preserve singletons.

Related issues:

* https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/219
* https://github.com/sourmash-bio/sourmash/issues/2271
* https://github.com/sourmash-bio/sourmash/issues/700
* https://github.com/sourmash-bio/sourmash/issues/225
* https://github.com/sourmash-bio/sourmash/issues/274


bluegenes:
@mr-eyes @ctb ready for review

ctb:
Related issues (might want to add into PR description @bluegenes) -

* https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/219
* https://github.com/sourmash-bio/sourmash/issues/2271
* https://github.com/sourmash-bio/sourmash/issues/700
* https://github.com/sourmash-bio/sourmash/issues/225
* https://github.com/sourmash-bio/sourmash/issues/274

bluegenes:
> Curious about the column name changes - why switch to cANI? (It's fine just curious.)

main reason is because I wanted the columns to be shorter `average_containment_ani` --> `average_ani`or`average_cANI`, and I was (maybe overly) concerned about being descriptive. It would actually be much simpler to just say `average_ani` and `max_ani`. Thoughts?

ctb:
> > Curious about the column name changes - why switch to cANI? (It's fine just curious.)
> 
> main reason is because I wanted the columns to be shorter `average_containment_ani` --> `average_ani`or`average_cANI`, and I was (maybe overly) concerned about being descriptive. It would actually be much simpler to just say `average_ani` and `max_ani`. Thoughts?

main important thing (IMO) is to try to standardize across our packages, including sourmash :).

ctb:
> > > There is still one difference -- in branchwater we are reporting ANI as percent (*100), whereas in sourmash we report it as a fraction. This is something I'd like to propose changing in sourmash at some point, but I could go back to fraction to keep it standardized.
> 
> > please do 😆
> 
> done, sigh.

😭 much appreciated

> > Is there a specific reason for representing ANI with 0-100% instead of 0-1 fractions?
> 
> more biologist friendly and similar to what many other tools report. e.g. when outputting kraken report format in sourmash, i convert to a percent to keep as close to possible as the original format.

I don't think we need to be particularly biologist friendly in our raw CSVs, tho. Syntactic and semantic mismatches are a huge problem for our own internal toolset tho!

mr-eyes:
> more biologist friendly and similar to what many other tools report. e.g. when outputting kraken report format in sourmash, i convert to a percent to keep as close to possible as the original format.

I am afraid this will introduce inconsistencies/confusion in handling the output files. We will need to take care of this information when doing any downstream processing.
In this PR, you are using ANI as fractions, not percentages. Even if reported as percentages, they are still required to follow the fraction format when doing a `cutoff` on creating the `edges`.

bluegenes:
> In this PR, you are using ANI as fractions, not percentages. Even if reported as percentages, they are still required to follow the fraction format when doing the user-defined CLI cutoff on creating the edges.

I previously (in ANI PR) had it all in percentage, including the CLI. No matter, it is all fractions now :). 

minor regret about my initial decision when I added ANI to sourmash 🤷🏻‍♀️

ctb:
> minor regret about my initial decision when I added ANI to sourmash 🤷🏻‍♀️

lol well if we're going to talk about regrets for early design decisions I'm sure I have a list somewhere...

ctb:
I ran `pairwise` and `cluster` just now. Ran into a few hiccups, fixed them here: https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/245

A few comments/questions/requests:

* could you add a test for `-h`?
* could you add a test that uses the default column name (so, tests for the other problem I found 😉 )
* could you add some minimal benchmarks (time/memory) for a standard-ish comparison, e.g. gtdb-reps, so that users know what to expect from both `pairwise` and `cluster` for a real-ish analysis? ISTR it's pretty fast against gtdb-reps.  (It's OK to punt this one to an issue for later.)

Other thoughts:

* it surprised me that `pairwise` does not output self-matches, but maybe that's ok?
* the downstream consequence of that was that the `cluster` output only contained the vast minority of sketch names! which was the real surprise. Might be mentioned in the documentation.
* it is surprising to me that `--cluster-sizes` is required. What do you think about making it optional?

## What I ran

```
sourmash scripts pairwise podar-ref.zip -o podar-ref-cmp.csv --ani
sourmash scripts cluster podar-ref-cmp.csv -o podar-ref-cmp.cluster.csv --cluster-sizes podar-ref-cmp.cluster.hist.csv
```

ctb:
(and, I mean, y'know - nice work! :)

bluegenes:
> the downstream consequence of that was that the cluster output only contained the vast minority of sketch names! which was the real surprise. Might be mentioned in the documentation.

This was initially confusing to me, as cluster keeps all nodes it sees (only doesn't add edges if they don't meet threshold). But of course, sketches with no similarity at all to other sketches will not appear in pairwise at all. It is probably worth evaluating the performance of writing self similarity to make cluster output robust. Otherwise I would probably default to using `multisearch` for clustering instead -- ideal scenario is to have all sketch names in the output!

ctb:
> > the downstream consequence of that was that the cluster output only contained the vast minority of sketch names! which was the real surprise. Might be mentioned in the documentation.
> 
> This was initially confusing to me, as cluster keeps all nodes it sees (only doesn't add edges if they don't meet threshold). But of course, sketches with no similarity at all to other sketches will not appear in pairwise at all. It is probably worth evaluating the performance of writing self similarity to make cluster output robust. Otherwise I would probably default to using `multisearch` for clustering instead -- ideal scenario is to have all sketch names in the output!

exactly! 😄 

This is why I think a note in the docs (either in pairwise, or in cluster, or both?) is a good idea. If it is a persistent confusion we can always add an option to pairwise.

ctb:
I think this is ready to go, right?

ctb (approved):
⭐

bluegenes:
just need to run the benchmarking!

ctb:
> just need to run the benchmarking!

you can punt to an issue... ;)

bluegenes:
> you can punt to an issue... ;)

will do!

Issue #40, opened 2023-08-23
enable CTRL-C in rust code?


ctb:

bluegenes:
+1!

bluegenes:
if reading zips, need to make sure temp_dir gets properly deleted. It should happen automatically when the variable goes out of scope, but maybe include a `temp_dir.close` just in case?

bluegenes:
so there's a much simpler way via pyo3! I just enabled this in the `directsketch` plugin: 

1. Pass in `py` /`_py` from `lib` to main function
- https://github.com/sourmash-bio/sourmash_plugin_directsketch/blob/main/src/lib.rs#L86-L91
- https://github.com/sourmash-bio/sourmash_plugin_directsketch/blob/main/src/directsketch.rs#L603
2. call `py.check_signals` in main function:
- https://github.com/sourmash-bio/sourmash_plugin_directsketch/blob/main/src/directsketch.rs#L669

ccbaumler:
While running `snakemake --profile`, the SLURM jobs running rust code like `manysketch` or `directsketch` will continue after the workflow is stopped from a workflow error. 
The `config.yaml` I was using may be found here -> https://github.com/dib-lab/portable-snakemake-workflows/blob/main/dotfiles/README.md
```
$ snakemake --version
7.32.4
```

ctb:
my guess is that this behavior is related to Rust multithreaded apps not responding to signals properly, which is why I suggested colton post it here ;)

ctb:
dup #602

PR #253, opened 2024-02-28
WIP: add Ctrl-C interrupt

bluegenes:
addresses https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/40

This PR introduces a `ThreadManager` struct that contains a syncsender, a writer thread, and an atomic bool, `interrupted`, which is set to `true` if `Ctrl-C` is detected. 

```
pub struct ThreadManager<T: Send + 'static> {
    pub sender: Option<SyncSender<T>>,
    pub writer_thread: Option<std::thread::JoinHandle<Result<()>>>,
    pub interrupted: Arc<AtomicBool>,
}
```

`ThreadManager` has three functions:
- `new` for building a new instance and setting up `Ctrl-C` handling
- `send` for sending data to the writer thread
- `perform_cleanup`, to close the threads. This standardizes cleanup across interruption and successful completion.

In each main function, we instantiate a `ThreadManager` and wrap in in `Arc<Mutex>` for safe sharing across threads. Then, while iterating through e.g. queries or against signatures, we check the status of the `interrupted` bool each iteration. If `interrupted` is ever true, we return, foregoing the remaining iterations. 

Sometimes the return is faster than others -- it really depends on how long each iteration takes (aka how often we are checking for interruption).

This PR adds this `ctrl-c` capture to:
- `pairwise`
- `manysearch`
- `manysketch`
- `multisearch`
- rocksdb `fastmultigather`
- rocksdb `manysearch`

I want to punt the remaining commands to an issue for future work, since they require some refactoring (`fastgather`/`fastmultigather`/`cluster` do not use `send`/`recv`) or it wouldn't even really be useful (`index`, `check`) since we only provide a super lightweight wrapper around core functionality.

Notes/Questions:
- Pytest has its own Ctrl-C handler and we can't have two. So we only set up the Ctrl-C handling if we are _not_ running with pytest.
- How do I properly test this, other than just running?
- I have no idea if there are better ways to do this, but it seems logical to me and seems to work :) 

**In progress: benchmark to ensure this doesn't inflate runtimes.**

## 12k ICTV viral genomes, scaled=200

79.8m comparisons

| command | version | time |
| -------- | -------- | -------- |
| pairwise |  PR  | 17s     |
| pairwise | v0.9.1    | 20s     | 
| multisearch | PR  | 32s |
| multisearch | v0.9.1    | ?|
| manysketch | PR | 4 s |
| manysketch | v0.9.1 | ? |
| manysearch | PR | ? |
| manysearch | v0.9.1 | ? |
| rocksdb manysearch | PR | ? |
| rocksdb manysearch | v0.9.1 | ? |
| rocksdb fastmultigather | PR | ? |
| rocksdb fastmultigather | v0.9.1 | ? |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -463,6 +463,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctrlc"
+version = "3.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b467862cc8610ca6fc9a1532d7777cee0804e678ab45410897b9396495994a0b"
+dependencies = [
+ "nix",
+ "windows-sys 0.52.0",
+]
+
+[[package]]
 name = "difflib"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -953,6 +963,17 @@ dependencies = [
  "thiserror",
  "xz2",
  "zstd",
+]
+
+[[package]]
+name = "nix"
+version = "0.27.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2eb04e9c688eff1c89d72b407f168cf79bb9e867a9d3323ed6c01519eb9cc053"
+dependencies = [
+ "bitflags 2.4.2",
+ "cfg-if",
+ "libc",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


ctb:
see https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/40#issuecomme
nt-2105327466

PR #256, opened 2024-02-28
MRG: add ANI to manysearch

bluegenes:
- adds `query_containment_ani`, `match_containment_ani`, `average_containment_ani`, `max_containment_ani` to `manysearch`
- just adds `query_containment_ani` to rocksdb `manysearch`, since we don't have access to target sig for other calculations.


bluegenes:
(ready for review @ctb)

PR #259, opened 2024-02-29
WIP: use identical column names as sourmash gather

bluegenes:
It came up in discussions for #234 that we should try to keep column names the same as in base sourmash.

This PR updates column names from rocksdb `fastmultigather` so that they match sourmash gather.

- `match_name` --> `name`
- `match_md5` --> `md5`
- `match_filename` --> `filename`

I'd love for us to update these in sourmash `gather` at some point, but this is not really the place to do that.


ctb:
dangit, taking me at my word and all...

We could standardize on the prefetch column names, which I think make a lot of sense.

bluegenes:
> We could standardize on the prefetch column names, which I think make a lot of sense.

prefetch cols, from https://github.com/sourmash-bio/sourmash/blob/latest/src/sourmash/search.py#L366C5-L390C26:
> prefetch_write_cols = [
        "intersect_bp",
        "jaccard",
        "max_containment",
        "f_query_match",
        "f_match_query",
        "match_filename",
        "match_name",
        "match_md5",
        "match_bp",
        "query_filename",
        "query_name",
        "query_md5",
        "query_bp",
        "ksize",
        "moltype",
        "scaled",
        "query_n_hashes",
        "query_abundance",
        "query_containment_ani",
        "match_containment_ani",
        "average_containment_ani",
        "max_containment_ani",
        "potential_false_negative",
    ]  #'match_abundance'

Sure. This introduces a minor annoyance: any gather <--> fastmultigather file comparisons will need to account for this name difference.

ctb:
true. But, eventually, we will move the fields over to the prefetch names as much as possible, per https://github.com/sourmash-bio/sourmash/issues/1555

bluegenes:
okay! i shall close :)

Issue #207, opened 2024-02-10
index:: do not write output if no signatures to write?


bluegenes:
As of #134, we write empty index files in cases where there were no signatures to load. This matches sourmash underlying sourmash index functionality, do we want to keep it, or write no file?

As of #197, I raise an error for this situation. Other than indexing, I standardized most code to exit gracefully when no signatures are loaded from either query or search files.

bluegenes:
In #197, I introduced a bool param for each function, `allow_failed_sigpaths`, which is used within `utils::report_on_collection_loading` to decide whether or not to allow loading failures. I set that to `false` for `index`, so that all signatures must load properly in order to index.

Where this fix does _not_ solve indexing woes, is when all signatures load correctly, but do not match desired parameters. This yields an empty collection (still a valid collection), which is then passed to `Revindex` for indexing.

in `utils::report_on_collection_loading`, we do check for an empty collection, but currently, we do the following:
```
if collection.is_empty() {
    eprintln!("No {} signatures loaded, exiting.", report_type);
    return Ok(());
}
```

This means we actually would need to check for empty collection within each function, as I've proposed in #267 for index. Also, it's actually a bit confusing, because we write `exiting`, but don't initiate exit.

I'm thinking I should perhaps return an error from `report_on_collection_loading` instead, which would cause all of the functions to exit based on an empty collection.

We would need to handle this differently if we allow reading multiple collections in -- we would only want to exit if we have 0 signatures to use.

ctb:
hah! we confronted a similar challenge a few years back in sourmash - https://github.com/sourmash-bio/sourmash/issues/1426. At least this reminded me of that.

bluegenes:
> hah! we confronted a similar challenge a few years back in sourmash - https://github.com/sourmash-bio/sourmash/issues/1426. At least this reminded me of that.
...so we did!

so do you have a vote on erroring out when we have 0 sigs to search?

> I'm thinking I should perhaps return an error from report_on_collection_loading instead, which would cause all of the functions to exit based on an empty collection. This would mean they error out when unable to load any signatures for either 'query', 'against', or general 'siglist'.

PR #267, opened 2024-03-07
MRG: index should error out if no sigs to index

bluegenes:
`index` has been failing silently if there are no signatures to index by producing a rocksdb database that is just empty of sigs. It's been quite frustrating with issues arising from protein databases with pre-0.9.0 (aka incorrect) manifests :).

This PR changes `index` to Error out if there are no sigs to index.

I thought this had been handled as per https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/207, but regardless, it is fixed here.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -20,6 +20,13 @@ pub fn index<P: AsRef<Path>>(
         allow_failed_sigpaths,
     )?;
 
+    if collection.len() == 0 {
+        bail!(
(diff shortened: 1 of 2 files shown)


ctb:
Might be good to track down why #197 didn't fix it? :)

ctb (approved):

bluegenes:
I think I'll merge this and then leave the larger changes to a different PR

PR #282, opened 2024-03-20
MRG: fail-exit when 0 signatures are loaded from a collection

### LINKED ISSUE #207: index:: do not write output if no signatures to write?

bluegenes:
Error out when 0 signatures are loaded from a collection. 

Since we currently only load a single collection as our query or database, 0 sigs in either means that no useful results would be obtained.

- Fixes #207

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -20,13 +20,6 @@ pub fn index<P: AsRef<Path>>(
         allow_failed_sigpaths,
     )?;
 
-    if collection.len() == 0 {
-        bail!(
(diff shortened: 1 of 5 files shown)


bluegenes:
@ctb ready for review!

ctb (approved):

PR #284, opened 2024-03-20
MRG: fix manysketch naming bug

### LINKED ISSUE #283: `manysketch` does not set name properly for multiple sketches per entry
naming bug likely introduced in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/184

`set_name` likely not working as intended. I think it sets the name for the first signature (e.g. k=21) but not for any subsequent (e.g. k=31,51).
```
 } else if !set_name {
      sig.set_name(name);
   // sourmash sets filename to last filename if merging fastas
      sig.set_filename(last_filename.as_str());
     set_name = true;
```

bluegenes:
- fixes #282

This bug caused sketch names to be set only for the first sketch for a given file, and _not_ set for any additional sketches. This means, if we were sketching at `k=21,k=31,k=51`, only the `k=21` sketches would have the name properly set.

### DIFF
--- a/src/manysketch.rs
+++ b/src/manysketch.rs
@@ -226,7 +226,6 @@ pub fn manysketch(
                                     sig.set_name(name);
                                     // sourmash sets filename to last filename if merging fastas
                                     sig.set_filename(last_filename.as_str());
-                                    set_name = true;
                                 };
                                 if moltype == "protein" {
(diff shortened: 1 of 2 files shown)


bluegenes:
@ctb ready for review

PR #293, opened 2024-03-29
WIP: optionally output graph in pajek format

bluegenes:
In `cluster`, we use `connected_components` to output a set of clusters. However, there are a number of other community detection algorithms we might be interested in using.

By outputting the network graph, we can read the raw graph into another program (networkx/igraph) and use community detection (e.g. `igraph.community_infomap`). This would be an alternative to using the `connected_components` clusters.

### DIFF
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -85,6 +85,7 @@ pub fn cluster(
     similarity_column: String,
     similarity_threshold: f64,
     cluster_sizes: Option<String>,
+    output_graph: Option<String>,
 ) -> Result<()> {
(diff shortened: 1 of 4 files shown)


ctb:
> If you are interested in interop with igraph, I recommend GraphML or GML, both of which have clear specifications and are intended for general purpose graph storage and exchange. At igraph, our goal with the Pajek reader/writer is interop specifically with the Pajek software, not with other systems.

cool! I've had some small experience with GML in spacegraphcats, it's pretty straightforward to emit.

PR #302, opened 2024-04-19
MRG: add abund estimation to `manysearch`

bluegenes:
Adds abundance columns `average_abund`, `median_abund`, `std_abund` to `manysearch` output. 

Changed behavior:
- Modifies `manysearch` to only build & write a result if we find an overlap between the query and search sigs. This was to avoid issues with abundance-related columns (average, median, std deviation calculations), but I think generally this:
- shortens the results file to only relevant results and 
- speeds things up a tad, since we stop doing any unnecessary calculations.

To Do:
- [x] check vs sourmash/ mgmanysearch to ensure abund estimation is correct

Unfortunately, I can only estimate abundance info from non-rocksdb `manysearch`. Abund estimation for rocksdb databases would require us to estimate these values w/in sourmash, since we don't have access to _which_ hashes match.

Note: for `mgmanysearch` type applications, the default threshold, 0.01` is far too stringent. Recommend setting threshold to 0 for these applications.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -4,43 +4,30 @@ on:
   push:
     branches: [latest]
 jobs:
-  mamba:
+  tests_on_linux:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
 
-    - name: Install latest stable rust
-      uses: actions-rs/toolchain@v1
-      with:
-         toolchain: stable
-         override: true
-         components: rustfmt, clippy
-
-    - name: Run cargo fmt
-      uses: actions-rs/cargo@v1
-      with:
-        command: fmt
-        args: --all -- --check
-
-    - name: cache rust
-      uses: Swatinem/rust-cache@v2
-
     - name: cache conda
       uses: actions/cache@v4
       env:
-        CACHE_NUMBER: 0
+        CACHE_NUMBER: 1
       with:
         path: ~/conda_pkgs_dir
         key:
           ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('environment.yml') }}
 
+    - name: cache rust
+      uses: Swatinem/rust-cache@v2
+
     - name: setup conda
-      uses: conda-incubator/setup-miniconda@030178870c779d9e5e1b4e563269f3aa69b04081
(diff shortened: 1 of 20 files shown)


ctb:
wow, that's a fantastic speed improvement 😆

ctb:
per @bluegenes -
>it's done - I was just waiting on some verification numbers from annie and lost track of it.

I will dig in!

ctb:
(I will probably try to do some validation on this along with https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/363 and #366! I'm on a roll!)

bluegenes:
@ctb We also had a couple other thoughts going through this. I think they were:

1. Would it be worth adding an `--abundance-from` param so the abundance could be from either query or search sigs?
2. could we offer a different threshold too, e.g. minimum number of hashes? For very short query sequences, our default % threshold is far too stringent.

ctb:
> @ctb We also had a couple other thoughts going through this. I think they were:
> 
>1. Would it be worth adding an `--abundance-from` param so the abundance could be from either query or search sigs?

Yes! If you get to it before I review and merge, great. If not, I'll create an issue 😉 
> 
>2. could we offer a different threshold too, e.g. minimum number of hashes? For very short query sequences, our default % threshold is far too stringent.

Sure! Sensitive default => better here.

ctb:
ran https://hackmd.io/tKpLr1ISR9mqHHmZbEvcow?view with both `mgmanysearch` and this PR.

executed:
```
ls -1 f_prausnitzii.sig.gz b_uniformis.sig.gz > query.txt
ls -1 CD*.sig.gz > against.txt
sourmash scripts manysearch against.txt query.txt -o search2.csv -t 0
```

and then loaded both CSVs in and explored.

Screenshot 2024-07-01 at 3 50 33 PM

The numbers look the same, so 🎉 !!

A few interesting differences -
* for this PR/`manysearch`, query is metagenomes, against is genomes; for `mgmanysearch`, the reverse;
* the fraction of metagenome covered is not available from this PR (b/c of speed, so ✅ )
* rows with 0s are not reported in this PR/`manysearch`, which again, fine!

I want to take a gander at the ANI values, but if those are fine (which I expect they will be :glare:) then I will merge, unless there is a last minute objection.

ctb:
🎉 they are identical. merging!

ctb (approved):

Issue #285, opened 2024-03-22
`fastmultigather` doesn't properly report # of sigs searched


ctb:
```
gathering all sketches in '../podar-ref-singleton.zip' against 'podar-ref.rocksdb' using 8 threads
Loaded DB
Reading query(s) from: '../podar-ref-singleton.zip'
Loaded 490 query signature(s)
DONE. Processed 0 search sigs
...fastmultigather is done!
```
Number of search sigs should be larger than 0 :)

Issue #299, opened 2024-04-12
Can't get fastmultigather to write to a specific output file


AnneliektH:
Running fastmultigather on a set of metagenome reads x a set of zip files and i want the output to be named as metagenome.x.zip.csv

Running as:
`sourmash scripts fastmultigather ../../sourmash_sketches/sketch_reads/ERR1135178.k21.zip ../../sourmash_sketches/split_sig/ERR11351_1.zip --output ERR1135178xERR11351_1.csv -k 21 -t 1000 -s 100 -c 1`

Gives me 2 output files:  `ERR1135178.gather.csv` and `ERR1135178.gather.csv`, which is not what I want

ctb:
right, see https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/239 - `--output` only works with a rocksdb. We have two quite different modes of operation of fastmultigather that need to converge at some point :).

AnneliektH:
Ahh too bad. Would there be a way to work around this using zip files? as i currently cannot get rockdbs to work

bluegenes:
@ctb are you ok with switching to single-file output for all queries? Or should we aim to support both?

ctb:
> @ctb are you ok with switching to single-file output for all queries? Or should we aim to support both?

Two hot takes -

I'm kind of against switching things up wilhe-nilhe just because? The branchwater plugin is becoming a bit of a UX nightmare already and changing up what we do in the middle of a point release seems suboptimal. We've done it before, we'll do it again, but I question the necessity here ;). Can't we just give @amterhorst a csvtk command that aggregates all of the CSV files into one or something??

I would be more supportive of making it so that _when_ `-o` is give, it aggregates all the output into a single file, but when it is _not_ given, it sticks with the current behavior. I fear that will lead to a lot of code nastiness.

This is all in the background of a complicated set of design decisions discussed over in https://github.com/sourmash-bio/sourmash/issues/2328, for which I have no answer.

So, umm, after all of this - yeah, sure, switch to `-o` aggregating all output into a single file. It'll be a required parameter now, right?

AnneliektH:
Ok: rocksdb not working was my fault, as i tried to create directly from a sig.gz file, not a list of filepaths.
So I can create output files with a specific name now

As for a csvtk command, that be great, but my issue is that the output file gets overwritten. 
Im trying to run the same metagenome against different sets of databases, and output files are named after the metag. 

[...]

Issue #239, opened 2024-02-26
`fastmultigather` `--output` is not being used


mr-eyes:
I know this is being used in `mastiff_manygather`, but I expected it to be used when I did the `fastmultigather` on sigs.

https://github.com/sourmash-bio/sourmash_plugin_branchwater/blob/43caeba6833a7efbdacc98b3cfbf56ae8867110e/src/python/sourmash_plugin_branchwater/__init__.py#L152

ctb:
I don't think it's being used there, either? Should probably just be removed entirely.

bluegenes:
It's being used in the rocksdb version!

The reason it's not also used in the non-rocksdb version is that that version produces more than one file (prefetch and gather).

This could certainly be harmonized better (or at least documented in the help).

bluegenes:
ref discussion around `multigather` output in sourmash:

- https://github.com/sourmash-bio/sourmash/issues/1614
- https://github.com/sourmash-bio/sourmash/issues/2328
- https://github.com/sourmash-bio/sourmash/pull/2065

PR #320, opened 2024-05-04
MRG: fix lint error, & processed sig reporting for rocksdb/fastmultigather

ctb:
Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/313
Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/285

Addresses #239 / #299 by bailing out when an output path is provided but will not be used:
```
RuntimeError: output path specified, but not running fastmultigather against a rocksdb. See 
issue #239
```
although we are leaning towards respecting `-o` in the future, see #299 discussion, so this will soon get revisited.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -12,14 +12,17 @@ jobs:
         fetch-depth: 0
 
     - name: Install latest stable rust
-      uses: dtolnay/rust-toolchain@stable
+      uses: actions-rs/toolchain@v1
       with:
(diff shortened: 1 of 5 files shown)


ctb:
@bluegenes I think this should be merged, despite CI tests not running - it compiles and passes tests on my laptop just fine, and the changes are easy to inspect.

bluegenes (approved):
lgtm

Issue #347, opened 2024-06-07
`index` doesn't work with a text file list of manifests


olgabot:
Hello, hope you are well!

I am very excited to try out the low-memory and fast searches created by RocksDB :) (Also, I will definitely be making use of `pairwise`!)

On my way there, I encountered some unexpected behavior. I had an enormous sequence file (e.g. [UniRef50, 65M protein sequences](https://www.uniprot.org/uniref?query=*&facets=identity%3A0.5)) and cut it up into chunks of 100k sequences to do `sourmash scripts manysketch -p protein,scaled=1,k=10,abund` without running out of resources.

Then, I wanted to index these many files before searching them, but `sourmash scripts index` didn't work on a list of manifest files.

Here's a minimal reproduction, using the data in `src/python/tests/test-data`:

```
# Make input csv files
echo 'name,genome_filename,protein_filename\nshort,short.fa,' > short.csv 
echo 'name,genome_filename,protein_filename\nshort,short2.fa,' > short2.csv
echo 'name,genome_filename,protein_filename\nshort,short3.fa,' > short3.csv

# Make sketches
sourmash scripts manysketch short.csv -o short.fa.zip -p dna,k=31,scaled=1 
sourmash scripts manysketch short2.csv -o short2.fa.zip -p dna,k=31,scaled=1
sourmash scripts manysketch short3.csv -o short3.fa.zip -p dna,k=31,scaled=1

# Make list of sketches (but they're actually manifests?)
for ZIP in short*.zip; do echo $ZIP >> short_siglist.txt; done
```

Then, `sourmash scripts index` fails

```
$ sourmash scripts index --ksize 31 --scaled 1 -o short_index.rocksdb short_siglist.txt   

== This is sourmash version 4.8.8. ==
== Please cite Brown and Irber (2016), doi:10.21105/joss.00027. ==

ksize: 31 / scaled: 1 / moltype: DNA 
indexing all sketches in 'short_siglist.txt'
Loading siglist
Reading signature(s) from: 'short_siglist.txt'
Sketch loading error: expected value at line 1 column 1
WARNING: could not load sketches from path 'short2.fa.zip'
Sketch loading error: expected value at line 1 column 1
WARNING: could not load sketches from path 'short.fa.zip'
Sketch loading error: expected value at line 1 column 1
WARNING: could not load sketches from path 'short3.fa.zip'
No valid signatures found in signature pathlist 'short_siglist.txt'
[...]

Issue #266, opened 2024-03-05
support standalone manifests containing zip files


ctb:
build a manifest containing zip files with `sig collect`:
```
sourmash sig collect podar-ref/1.fa.sig.zip -o manifest-with-zipfiles.csv -F csv
```

try to run manysearch on it:
```
sourmash scripts manysearch manifest-with-zipfiles.csv manifest-with-zipfiles.csv -o xxx.csv
```

and you will get:
```
Reading query(s) from: 'manifest-with-zipfiles.csv'
Loaded 1 query signature(s)
thread '<unnamed>' panicked at src/manysearch.rs:32:90:
called `Result::unwrap()` on an `Err` value: Error: Failed to load query record: CP001941.1 Aciduliprofundum boonei T469, complete genome
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/Users/t/miniforge3/envs/py311/bin/sourmash", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/t/dev/sourmash/src/sourmash/__main__.py", line 20, in main
    retval = mainmethod(args)
             ^^^^^^^^^^^^^^^^
  File "/Users/t/dev/pyo3_branchwater/src/python/sourmash_plugin_branchwater/__init__.py", line 71, in main
    status = sourmash_plugin_branchwater.do_manysearch(args.query_paths,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: Error: Failed to load query record: CP001941.1 Aciduliprofundum boonei T469, complete genome
```
Also note that we should load files from within manifests as if they are relative to the manifest dir per https://github.com/sourmash-bio/sourmash/pull/3054 and https://github.com/sourmash-bio/sourmash/issues/3008#issuecomment-1975174211.

Originally noted in https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/237.

ctb:
note: #364 temporarily removes the text from the docs that suggests using manifest CSVs.

ctb:
I dug into this some more today. The chain of code is:

In `utils.rs`, `collection_from_manifest` loads the manifest as a `Collection`:

[...]

Issue #235, opened 2024-02-22
recommend (and test) 'standalone manifests'


ctb:
related to https://github.com/sourmash-bio/sourmash/issues/3023, I think we should proselytize the use of _standalone_ manifests in certain circumstances.

We refer to 'manifest CSVs' in the current docs, but this is confusing, since it turns out manifest CSVs generated by `sig manifest` do not necessarily work as standalone manifests, while manifests generated by `sig collect` do. Standalone manifests are (in my newspeak) files that can be loaded & used by sourmash as, well, standalone files - they are not tightly linked to a specific storage type.

I'm going to work on the framing and documentation and tests a bit over in the sourmash repo, but one of the main goals is to get it straight for use in the branchwater plugin 😅 so I thought I'd create an issue here to make sure we update the docs appropriately (and make sure the functionality actually works).

ctb:
probably should link to https://sourmash.readthedocs.io/en/latest/databases-advanced.html in the docs here, as well.

ctb:
The documentation is being significantly reworked in https://github.com/sourmash-bio/sourmash/pull/3054 and https://github.com/sourmash-bio/sourmash/pull/3027. Once those are merged, we can update here.

In brief, we should recommend using `sig collect` and/or `sig check` to build standalone manifests.

ctb:
Support for this is being added in #430.

Issue #264, opened 2024-03-05
provide a better error when trying to read a zip file without a manifest


ctb:
When we run:
```
scripts manysearch zip-with-manifest.zip zip-with-manifest.zip -o foo.csv
```
it works fine; but with
```
sourmash scripts manysearch zip-without-manifest.zip zip-without-manifest.zip -o foo.csv
```
we get
```
Reading query(s) from: 'zip-without-manifest.zip'
Sketch loading error: No such file or directory (os error 2)
WARNING: could not load sketches from path 'PK'
No valid signatures found in query pathlist 'zip-without-manifest.zip'
WARNING: 1 query paths failed to load. See error messages above.
No query signatures loaded, exiting.
Reading search(s) from: 'zip-without-manifest.zip'
Sketch loading error: No such file or directory (os error 2)
WARNING: could not load sketches from path 'PK'
No valid signatures found in search pathlist 'zip-without-manifest.zip'
WARNING: 1 search paths failed to load. See error messages above.
No search signatures loaded, exiting.
DONE. Processed 0 search sigs
...manysearch is done! results in 'foo.csv'
```

It's fine that it doesn't work but it'd be nice to provide a better error message, like "this is a zip file without a manifest, you suxor."

Related issues:
* https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/237

ctb:
So, this seems a little weird - in utils.rs,
```rust
pub fn collection_from_zipfile(sigpath: &Path, report_type: &ReportType) -> Result<Collection> {
    match Collection::from_zipfile(sigpath) {
        Ok(collection) => Ok(collection),
        Err(_) => bail!("failed to load {} zipfile: '{}'", report_type, sigpath\
),
    }
}
```
and then in sourmash collection.rs,
```rust
    pub fn from_zipfile<P: AsRef<Path>>(zipfile: P) -> Result<Self> {
        let storage = ZipStorage::from_file(zipfile)?;
        // Load manifest from standard location in zipstorage                   
        let manifest = Manifest::from_reader(storage.load("SOURMASH-MANIFEST.csv")?.as_slice())?;
        Ok(Self {
            manifest,
            storage: InnerStorage::new(storage),
        })
    }
```

and it would seem to me that `storage.load` should fail if `SOURMASH-MANIFEST.csv` doesn't exist in the zip file, and the error should be propogated.

ctb:
[...]

PR #354, opened 2024-06-13
Index manifest zip files

olgabot:
For https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/347, https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/266, https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/235, this PR (will eventually) add the ability to do `sourmash scripts index` on multiple manifests at once.

### DIFF
--- a/src/python/tests/test_index.py
+++ b/src/python/tests/test_index.py
@@ -123,6 +123,32 @@ def test_index_manifest(runtmp, capfd):
     assert 'index is done' in runtmp.last_result.err
 
 
+def test_index_multiple_manifests(runtmp, capfd):
+    # test index with text file of multiple manifests
+    sig2 = get_test_data('2.fa.sig.gz')
+    sig47 = get_test_data('47.fa.sig.gz')
+    sig63 = get_test_data('63.fa.sig.gz')
+    sigs = [sig2, sig47, sig63]
+    manifests = []
+    for sig in sigs:
+        sig_mf = runtmp.output(os.path.basename(sig) + ".mf.zip")
+        runtmp.sourmash("sig", "cat", sig, "-o", sig_mf)
+        manifests.append(sig_mf)
+
+    # assert False
+    manifests_zips_list = runtmp.output('manifest_zips.txt')
+    make_file_list(manifests_zips_list, manifests)
+
+    output = runtmp.output('out.db')
+    runtmp.sourmash('scripts', 'index', manifests_zips_list,
+                        '-o', output)
+
+    captured = capfd.readouterr()
+    print(captured.err)
+    print(runtmp.last_result.err)
+    assert 'index is done' in runtmp.last_result.err
+
+
 def test_index_bad_siglist_2(runtmp, capfd):
     # test with a bad siglist (containing a missing file)
     against_list = runtmp.output('against.txt')


src/utils.rs
```diff
@@ -585,58 +625,50 @@ fn collection_from_pathlist(
...
+                Some(collection
+                    .unwrap()
+                    .manifest()
+                    .iter()
+                    .filter_map(|record| match record.internal_location(){
+                        PathBuf => Some(location)
+                        
+                    })
+                    .collect())
+            },
```
olgabot, line 657: Getting compiler error that `Some` doesn't exist? Not sure what to make of this..

```
error[E0425]: cannot find value `location` in this scope
   --> src/utils.rs:653:41
    |
653 |                         PathBuf => Some(location)
    |                                         ^^^^^^^^ not found in this scope

```

src/utils.rs
```diff
@@ -585,58 +625,47 @@ fn collection_from_pathlist(
...
+            Ok(collection) => {
+                // For each record in the collection, get its path filename
+                Some(collection
+                    .unwrap()
+                    .manifest()
+                    .iter()
+                    .map(|record| record.internal_location())
+                    .collect())
+            },
             Err(err) => {
```
olgabot, line 655: tried using `map` here, but then the type couldn't be inferred:

```rust
error[E0283]: type annotations needed
   --> src/utils.rs:653:22
    |
645 |         .filter_map(|path| match collection_from_zipfile_or_signature_or_manifest(&path, &report_type) {
    |          ---------- type must be known at this point
...
653 |                     .collect())
    |                      ^^^^^^^ cannot infer type of the type parameter `B` declared on the method `collect`
    |
    = note: cannot satisfy `_: Send`
note: required by a bound in `rayon::iter::ParallelIterator::filter_map`
   --> /Users/olgabot/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/mod.rs:840:12
    |
837 |     fn filter_map<P, R>(self, filter_op: P) -> FilterMap<Self, P>
    |        ---------- required by a bound in this associated function
...
840 |         R: Send,
    |            ^^^^ required by this bound in `ParallelIterator::filter_map`
help: consider specifying the generic argument
    |
653 |                     .collect::<Vec<_>>())
    |                             ++++++++++
```

ctb:
Hi @olgabot I can't quite tell what you're trying to do in this PR 😅 

The code in `test_index_multiple_manifests` is creating zip files, not manifests - each `.mf.zip` file is a zip file containing one sketch, right?

```
% sourmash sig cat src/python/tests/test-data/63.fa.sig.gz -o xyz.mf.zip
...
% unzip -v xyz.mf.zip
Archive:  xyz.mf.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
   43584  Stored    43584   0% 06-30-2024 16:38 903b1180  signatures/38729c6374925585db28916b82a6f513.sig.gz
     337  Defl:N      253  25% 06-30-2024 16:38 b1f33f83  SOURMASH-MANIFEST.csv
--------          -------  ---                            -------
   43921            43837   0%                            2 files
```

By contrast, over in #235, 
```
% sourmash sig collect src/python/tests/test-data/63.fa.sig.gz -o mf.csv -F csv  
...
% cat mf.csv
# SOURMASH-MANIFEST-VERSION: 1.0
internal_location,md5,md5short,ksize,moltype,num,scaled,n_hashes,with_abundance,name,filename
src/python/tests/test-data/63.fa.sig.gz,38729c6374925585db28916b82a6f513,38729c63,31,DNA,0,1000,5238,False,"NC_011665.1 Shewanella baltica OS223 plasmid pS22303, complete sequence",podar-ref/63.fa
```
creates a standalone manifest CSV file - a CSV file that contains a pointer to sketches in another file.

Revisiting your test, you're essentially doing this, right?
```
sourmash sig cat src/python/tests/test-data/63.fa.sig.gz -o 63.sig.zip
sourmash sig cat src/python/tests/test-data/47.fa.sig.gz -o 47.sig.zip
ls -1 47.sig.zip 63.sig.zip > list.txt
```

and then running things like `manysearch` on list.txt.

This also doesn't currently work (except maybe in this branch!) and it's great to get it working, but it's different from manifests.

ctb:
when I run manysearch on list.txt in the main branch, I get:

```
% sourmash scripts manysearch list.txt list.txt -o xyz.csv

== This is sourmash version 4.8.11.dev0. ==
== Please cite Irber et. al (2024), doi:10.21105/joss.06830. ==

=> sourmash_plugin_branchwater 0.9.6.dev0; cite Irber et al., doi: 10.1101/2022.11.02.514947

ksize: 31 / scaled: 1000 / moltype: DNA / threshold: 0.01
searching all sketches in 'list.txt' against 'list.txt' using 8 threads
selection scaled: Some(1000)
Reading query(s) from: 'list.txt'
Sketch loading error: expected value at line 1 column 1
WARNING: could not load sketches from path '47.sig.zip'
Sketch loading error: expected value at line 1 column 1
WARNING: could not load sketches from path '63.sig.zip'
No valid signatures found in query pathlist 'list.txt'
WARNING: 2 query paths failed to load. See error messages above.
Error: No query signatures loaded, exiting.
```

If that's what you want to get working, then a better PR title here would be "enable listing of zip files in fromfiles".

ctb:
(and I verify that under this PR, the above command works fine 🎉 )

ctb:
> > Hi @olgabot I can't quite tell what you're trying to do in this PR 😅
> 
> Yes, sorry, I got confused by manifests vs zip files vs signatures here 😅 . The goal of this PR is to allow for `sourmash scripts index` to work on a `pathlist.txt` containing zip files. The suggested title of "enable listing of zip files in fromfiles" doesn't make sense to me because the main affected method is `collection_from_pathlist` called by `load_collection`, and I don't know what "fromfiles" means, sorry. What about "Allow lists of zip files to work in `collection_from_pathlist`"?

heh, I almost had suggested `pathlist`.

We (I) have been inconsistent, it seems - the branchwater plugin documentation uses fromfile (e.g. [link](https://github.com/sourmash-bio/sourmash_plugin_branchwater/blob/main/doc/README.md#using-fromfiles)), while sourmash uses pathlist (e.g. see refactoring [here](https://github.com/sourmash-bio/sourmash/pull/3027)).

So anyway two comments -
* pathlists are suboptimal, because it requires two "reads" for each file - first, reading to get a manifest, and then again to actually load the sketch. Benchmarks [show](https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/214) it's not that big a deal in the end, at least not for small files.
* I would generally suggest trying to name PRs that improve command-line functionality after whatever terminology we use in the docs; this is because those titles end up in the release docs and presumably someone might actually read those. In practice this probably doesn't matter. But that's why I suggested 'fromfile' rather than basing things on the internal names. Given the sourmash inconsistencies above I'd suggest actually fixing the docs here to use pathlist but that's not really your problem to deal with!

tl;dr it would be good to fix the PR name/description somehow.

On a side note - 

This issue is also worth skimming for @bluegenes discussion of how the code works: https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/264

ctb:
hi @olgabot - not sure if this applies to this PR, but I wanted to alert you to some considerations that are coming up for manifests and relative path interpretation over in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/390#issuecomment-2225776967. Happy to chat more if you think it is relevant.

ctb:
(I think as long as you are working with fromfiles/pathlists you're fine, and the main recommendation for you currently is - get things working for absolute paths first, because relative path interpretation is a bit wibbly and might need some tender loving care in the future.)

ctb:
@olgabot I tackled this from another angle - standalone manifests - and I ended up writing a whole new struct, `MultiCollection`. Main PR here, https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/434, working to get it to pass all the tests here: https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/430. I think this will address most of what you set out to accomplish here, although `index` is proving to be a bit of a bear.

ctb:
> Hmm though it does seem like `index` may be working now from [sourmash-bio/sourmash#3250](https://github.com/sourmash-bio/sourmash/pull/3250) ... let me try that

(right, for internal storage it can work, but ... probably not yet ;)

ctb:
> Internal storage is fine for me since all the files are on s3 anyway, so referencing other blobs isn't the best idea. However, I'm still not able to do `sourmash index list_of_zips.txt` :( From what I can tell, I think that's what #430 is implementing with `MultiCollection`, is that right?

#430 will probably not implement this for `sourmash scripts index`, specifically, but will for everything else.

For now, `RevIndex` with non-internal storage requires a single `Storage` struct, which works out to a single zip file, sig.gz file, or other kind of supported backend.

Although, I didn't think about supporting it just for internal storage... that should actually work. Will see what I can do :).

ctb:
> For now, `RevIndex` with non-internal storage requires a single `Storage` struct, which works out to a single zip file, sig.gz file, or other kind of supported backend.
> 
> Although, I didn't think about supporting it just for internal storage... that should actually work. Will see what I can do :).

Implemented in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/451, which I'll be merging into #450 and #430. I'll update here when it's officially tested, but it _should_ work now.

PR #359, opened 2024-06-17
MRG: update ReportType string output

ctb:
I noticed a strange output when running `pairwise`:
```
Loaded 1000 signature signature(s)
```
I was curious about the redundancy and tracked this down to ReportType => string formatting.

I think having this be:

Loaded 1000 analysis signature(s)

makes a bit more sense.

Small change, but 🤷 :)


bluegenes (approved):
Lgtm! That is better 😂

PR #375, opened 2024-07-01
WIP: update `manysketch` `--singleton`

bluegenes:
I'm running into memory issues with `manysketch` `--singleton`, likely brought on by the fact that we're storing all the singleton sigs in a list rather than sending them to be written. This PR fixes that by writing singletons immediately before going on to the next read/contig. We then avoid writing singletons at the end of the loop, where they would just be empty template sigs.

### DIFF
--- a/src/manysketch.rs
+++ b/src/manysketch.rs
@@ -170,7 +170,6 @@ pub fn manysketch(
             let name = &fastadata.name;
             let filenames = &fastadata.paths;
             let moltype = &fastadata.input_type;
-            let mut allsigs = Vec::new();
             // build sig templates for these sketches from params, check if there are sigs to build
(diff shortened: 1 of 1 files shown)


src/manysketch.rs
```diff
@@ -243,20 +242,26 @@ pub fn manysketch(
...
+                            eprintln!("Unable to send internal data: {:?}", e);
+                            return None;
+                        }
                         sigs = sig_templates.clone();
                     }
                 }
             }
-            if !singleton {
-                allsigs.append(&mut sigs);
+            // singletons have already been written -> avoid writing empty templates
```
ctb, line 254:
```suggestion
            // if singleton sketches, they have already been written; only write aggregate sketches
```

ctb (approved):

PR #392, opened 2024-07-13
MRG: add error-exit in fastmultigather against rocksdb for failed gather

ctb:
This fixes an issue noted in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/390#issuecomment-2225942342, where if a RocksDB cannot load a sketch from external FSStorage, `fastmultigather` outputs an error message, but doesn't actually error exit.

With this PR, a clear error message is printed. However, we don't quit out of the fastmultigather early; that seemed like a harder fix :).


bluegenes (approved):
Looks good!

Issue #379, opened 2024-07-03
lower `manysearch` default threshold


ctb:
per https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/302#issuecomment-2183135247,

>could we offer a different threshold too, e.g. minimum number of hashes? For very short query sequences, our default % threshold is far too stringent.

ctb:
actually, no 😭 

per https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/388, query and target were flipped, which made 0.01 too stringent. Since we reversed this to the original meaning in #388, 0.01 refers to genome-in-metagenome, so 1% of the genome is present - which is good.

Life is pain.

Issue #377, opened 2024-07-03
script to compare `mgmanysearch` results with `manysearch`


ctb:
The following script verifies that the results returned by `mgmanysearch` (from `sourmash_plugin_containment_search`) match the results returned by `manysearch` when run from this branch.

This should be added as a test.

See https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/302#issuecomment-2201310187 for tutorial and runtime; also https://github.com/sourmash-bio/sourmash/issues/3237.

```python
import pandas as pd

py_results = 'search.csv'
rs_results = 'search2.csv'

pydf = pd.read_csv(py_results)
rsdf = pd.read_csv(rs_results)

rs_columns = ['query_name', 'query_md5', 'match_name', 'containment',
        'match_md5', 'jaccard',
       'average_abund', 'median_abund', 'std_abund', 'query_containment_ani',
       'match_containment_ani', 'average_containment_ani',
       'max_containment_ani'] # 'intersect_hashes',

rename_py_col = { 'query_name': 'match_name',
                  'match_name': 'query_name',
                  'query_md5': 'match_md5',
                  'match_md5': 'query_md5',
                  'genome_containment_ani': 'match_containment_ani',
                  'match_containment_ani': 'query_containment_ani',
                  'f_match': 'containment'}

rs_compare_df = rsdf[rs_columns]
rs_compare_df['std_abund'] = rs_compare_df['std_abund'].round(5)
rs_compare_df.set_index('query_name', inplace=True)

py_compare_df = pydf.rename(columns=rename_py_col)
py_compare_df = py_compare_df[rs_columns]
py_compare_df = py_compare_df[py_compare_df['containment'] > 0]
py_compare_df['std_abund'] = py_compare_df['std_abund'].round(5)
py_compare_df.set_index('query_name', inplace=True)

assert py_compare_df.equals(rs_compare_df)

print(f"after adjustments, '{py_results}' == '{rs_results}'")
```

ctb:
The update/fix in #388 makes this a bit simpler 😅 , which is nice. I was worrying about this in the back of my mind!

```python
# %load check.py
import pandas as pd

py_results = 'search.csv'
rs_results = 'search2.csv'

pydf = pd.read_csv(py_results)
rsdf = pd.read_csv(rs_results)

rs_columns = ['query_name', 'query_md5', 'match_name', 'containment',
        'match_md5', 'jaccard',
[...]

Issue #380, opened 2024-07-03
add metagenome weighted/unweighted size (and maybe other columns) to `manysearch`


ctb:
~With #302, we have all the difficult-to-implement columns in place for `manysearch` CSV output - including especially the abundance of the query in the metagenome. That having been said, there are a few columns we should consider adding.~

~I'm particularly interested in supporting "nice" output, as provided by [the containment_search plugin]() / per the containment search tutorial [link](https://github.com/sourmash-bio/sourmash/issues/3237):~

```
query             p_genome avg_abund   p_metag   metagenome name
--------          -------- ---------   -------   ---------------
F. prausnitzii       4.3%    24.7       2.5%     CD136
B. uniformis        48.3%     2.0       3.4%     CD136
F. prausnitzii       0.0%     0.0       0.0%     CD237
B. uniformis        62.6%    14.5      26.5%     CD237
```

~in particular, it looks like we cannot calculate `p_metag` from the information we have available in `manysearch` which is kind of a problem for some use cases 😅 . It might need to be an optional switch because it could slow down manysearch too.~ **NOTE: added in #388**

While we're at it, we might usefully add a few other columns like ksize, scaled, and so on. I'm mildly worried about the file becoming too big, but these are useful columns... not sure how to think about it.

here's a script that produces the `mgmanysearch` nice output from `manysearch` CSV output -

`display-manysearch.py`:
```python
#! /usr/bin/env python
import argparse
import sys
import csv


def main():
    p = argparse.ArgumentParser()
    p.add_argument('manysearch_csv')
    args = p.parse_args()

    with open(args.manysearch_csv, newline='') as fp:
        r = csv.DictReader(fp)
        rows = list(r)

    rows.sort(key=lambda row: row['query_name']) # sort by metagenome, for now

    first = True
    for row in rows:
        has_abundance = row['average_abund'] != ''

        #
        # display!
        #

        # displaying first result?
        if first:
            print("")
            print("query             p_genome avg_abund   p_metag   metagenome name")
            print("--------          -------- ---------   -------   ---------------")
            first = False

        f_genome_found = float(row['containment'])
        pct_genome = f"{f_genome_found*100:.1f}"

        if has_abundance:
            n_weighted_found = int(row['n_weighted_found'])
            total_weighted_hashes = int(row['total_weighted_hashes'])
            f_metag_weighted = n_weighted_found / total_weighted_hashes # results_d['f_match_weighted']
[...]

PR #408, opened 2024-08-11
MRG: add pretty printing option to manysearch command

ctb:
This PR adds an option`-P/--pretty-print` to `manysearch` that uses the code in https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/380 to produce nice output.

The default is to pretty-print, which is maybe not great if you're generating thousands of results. But it's easy to turn off with `-N`.

This PR _also_ adjusts `manysearch` so that the `average_abund`, `median_abund`, and `std_abund` are not present when abundance is not available, which follows practice elsewhere in the code base.

Questions:

* should we add this to `pairwise` a la `sourmash compare`, too?
* same question for `fastgather` a la `sourmash gather`...

Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/379
Closes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/377 because it adds an explicit test for the output.

TODO:
- [x] add tests


ctb:
@sourmash-bio/devs read for review & merge!

bluegenes (approved):
lgtm!

PR #424, opened 2024-08-17
MRG: misc cleanup

ctb:
Cleanup:
- make moltype "DNA" instead of "dna" to match sourmash

Refactoring:
- consolidate duplicate test functions
- move common parametrize marks from pytest into `conftest.py`
- rename `test_gather.py` to `test_fastgather.py`
- rename `test_multigather.py` to `test_fastmultigather.py`


ctb:
Ready for review and merge @bluegenes!

PR #446, opened 2024-08-29
MRG: optionally, create empty results files for in-memory `fastmultigather`

bluegenes:
In-memory `fastmultigather` currently only creates output files (`*.gather.csv`, `*.prefetch.csv` if matches are found. Here, we add an option `--create-empty-results` (already a standard `sourmash gather` option; added in https://github.com/sourmash-bio/sourmash/pull/2557 ) to touch the empty files. This is primarily useful for snakemake workflows, we often need all output files to be created for use in downstream steps.

Note: **this only creates empty files if we would expect results (i.e. matching sketch parameters but just no overlapping hashes).** We could also create empty files for sketches without matching params, but that was unnecessary for my application and seemed excessive to me. Happy to add if there's a good reason.

When running `sourmash tax` on empty gather files, we can use `--force` to continue past these files. If using just one gather file, `tax` will not produce empty results. However, we can then `touch` tax output files within the snakemake rule to avoid workflow issues.

### DIFF
--- a/src/python/tests/test_fastmultigather.py
+++ b/src/python/tests/test_fastmultigather.py
@@ -43,13 +43,9 @@ def test_simple(runtmp, zip_against):
     if zip_against:
         against_list = zip_siglist(runtmp, against_list, runtmp.output('against.zip'))
 
-    cwd = os.getcwd()
-    try:
-        os.chdir(runtmp.output(''))
-        runtmp.sourmash('scripts', 'fastmultigather', query_list, against_list,
-                        '-s', '100000', '-t', '0')
-    finally:
-        os.chdir(cwd)
+
+    runtmp.sourmash('scripts', 'fastmultigather', query_list, against_list,
+                        '-s', '100000', '-t', '0', in_directory=runtmp.output(''))
 
     print(os.listdir(runtmp.output('')))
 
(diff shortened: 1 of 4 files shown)


bluegenes:
@ctb - ready for review

src/python/tests/test_fastmultigather.py
```diff
@@ -1254,3 +1254,30 @@ def test_save_matches(runtmp):
...
+    sig47 = get_test_data('47.fa.sig.gz')
+    sig63 = get_test_data('63.fa.sig.gz')
+
+    query_list = runtmp.output('query.txt')
+    against_list = runtmp.output('against.txt')
+
+    make_file_list(query_list, [sig2])
+    make_file_list(against_list, [sig47, sig63])
+
+    cwd = os.getcwd()
```
ctb, line 1271: note that `runtmp.sourmash(..., in_directory=runtmp.output(''))` will do this getcwd/try/finally for you ;).

ctb (approved):
looks great! I have notes to go fix the getcwd/try/finally through the test_fastmultigather file, so you can do it here or not.

src/python/tests/test_fastmultigather.py
bluegenes: just fixed in cc9f89b0e5cf1d5367f2413ee837c1c3dd7c10cd

PR #458, opened 2024-09-28
Add probability of overlap and weighted containment for Multisearch matches

olgabot:
Originally explored in [this notebook](https://github.com/olgabot/botryllus-mhc/blob/main/notebooks/206b_e_values_per_kmer_logprob.ipynb), some high-containment hits are a result of highly frequent k-mers, and I want to downweight the containment by the probability of overlap. As I imagine it now, the frequencies would be computed based on all queries and all database signatures.

Here is a worked example, please let me know if I am missing something:

### query: `ACGTTTTT`

3-mers (6 total):

1. `ACG`
2. `CGT`
3. `GTT`
4. `TTT` x 3

### target: `TTTTTTTTTAC`

3-mers (8 total):

1. `TTT` x 7
2. `TAC`

### Containment:

intersecting k-mers in query / intersecting k-mers in target 

= 3/7

### Probability of overlap:

frequency of intersecting k-mers in query * frequency of intersecting k-mers in target

= 3/6 * 7/8

= 1/2 * 7/8

= 7/16

### Weighted containment:

Containment / Probability of overlap = Containment * (1/Probability of overlap)

= 3/7 * 7/16

= 3/16

Update: fix k-mers in example


ctb:
hi @olgabot thanks! I'll take a look when I can. Poke me in a day or two if I haven't gotten back.

In re black, it will complicate review a bit right now. I'm game for applying black to the repo, not sure the best way to get that done in the context of this PR. If you have a proposal, I'm game :).

(You could, for example, apply black in a separate PR, and then we'll merge that, and then review the new diff.)

ctb:
thanks for https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/492! Now that it's merged, I think you can:

1. apply black formatting to this PR *before* merging main
2. merge in main, and use something like `git restore --ours <conflicted file(s)>`  + `git add` to resolve conflicts
3. voila, fini

src/multisearch.rs
```diff
@@ -31,7 +39,7 @@ pub fn multisearch(
         allow_failed_sigpaths,
     )?;
 
-    let queries = query_collection.load_sketches(selection)?;
+    let queries: Vec<SmallSignature> = query_collection.load_sketches(selection)?;
```
ctb, line 42: why this change? I think it's fine to be explicit just curious :)

src/multisearch.rs
```diff
@@ -41,7 +49,63 @@ pub fn multisearch(
         allow_failed_sigpaths,
     )?;
 
-    let against = against_collection.load_sketches(selection)?;
+    let againsts: Vec<SmallSignature> = against_collection.load_sketches(selection)?;
+
+    let mut n_comparisons = 0.0;
+    // let mut queries_merged_mh: KmerMinHash = Default::default();
```
ctb, line 55: pls remove if no longer needed :)

src/multisearch.rs
```diff
@@ -41,7 +49,63 @@ pub fn multisearch(
...
+    // query_term_frequencies is md5 to hashval: freq mapping
+    let mut query_term_frequencies: HashMap<String, HashMap<u64, f64>> = Default::default();
+    let mut inverse_document_frequency: HashMap<u64, f64> = Default::default();
+
+    if estimate_prob_overlap {
+        n_comparisons = againsts.len() as f64 * queries.len() as f64;
+        // Combine all the queries and against into a single signature each, to get their
+        // underlying distribution of hashes across the whole input
+        eprintln!("Merging queries ...");
+        let queries_merged_mh: KmerMinHash = merge_all_minhashes(&queries).unwrap();
```
ctb, line 69: please use `expect("message")` instead of `unwrap` so we get a nicer error message.

src/multisearch.rs
```diff
@@ -41,7 +49,63 @@ pub fn multisearch(
...
+    // let mut queries_merged_mh: KmerMinHash = Default::default();
+    // let mut against_merged_mh: KmerMinHash = Default::default();
+    let mut query_merged_frequencies: HashMap<u64, f64> = Default::default();
+    let mut against_merged_frequencies: HashMap<u64, f64> = Default::default();
+
+    // query_term_frequencies is md5 to hashval: freq mapping
+    let mut query_term_frequencies: HashMap<String, HashMap<u64, f64>> = Default::default();
+    let mut inverse_document_frequency: HashMap<u64, f64> = Default::default();
+
+    if estimate_prob_overlap {
```
ctb, line 64: would this make more sense as a function? (the whole contents of the 'if' block, I mean)

src/multisearch.rs
```diff
@@ -86,6 +150,38 @@ pub fn multisearch(
...
                     let mut max_containment_ani = None;
+                    let mut prob_overlap: Option<f64> = None;
+                    let mut prob_overlap_adjusted: Option<f64> = None;
+                    let mut containment_adjusted: Option<f64> = None;
+                    let mut containment_adjusted_log10: Option<f64> = None;
+                    let mut tf_idf_score: Option<f64> = None;
+
+                    // Estimate probability of overlap between query sig and against sig, using
+                    // underlying distribution of hashvals for all queries and all againsts
+                    if estimate_prob_overlap {
```
ctb, line 161: maybe a function -> struct of values returned? So you do `Option<ProbStruct>` to encapsulate return?

src/python/sourmash_plugin_branchwater/__init__.py
```diff
@@ -456,6 +456,12 @@ def __init__(self, p):
         p.add_argument(
             "-a", "--ani", action="store_true", help="estimate ANI from containment"
         )
+        p.add_argument(
+            "-p",
+            "--prob",
```
ctb, line 461: suggest a longer long argument - it can be as verbose as you want (which is better for UX) but will still match on substrings like `--prob`.

ctb (changes requested):
preliminary review - looks pretty good! a few minor changes suggested, and maybe some cleanup -> refactoring into functions.

src/multisearch.rs
olgabot: I was getting really confused about types, so it was helpful for me to be explicit about everything. `rust-analyzer` in VS Code adds the types, but I wasn't able to tell what was there in the code vs what was being type-hinted by the IDE, so I decided to put them there explicitly.

ctb: ahh, gotcha! I have done the same in a few recent PRs :).

src/multisearch.rs
olgabot: Makes sense to me! Wrote out a separate `compute_single_prob_overlap` function

src/multisearch.rs
olgabot: Yep, created a `ProbOverlapStats` structure

src/multisearch.rs
olgabot: Done!

src/multisearch.rs
olgabot: Done!

src/python/tests/test_fastgather.py
```diff
@@ -461,7 +461,6 @@ def test_bad_against_3(runtmp, capfd):
     assert "InvalidArchive" in captured.err
 
 
-@pytest.mark.xfail(reason="should work, bug")
```
ctb, line 464: one of your merges must have removed this line by accident. Reinstate, it and you'll be passing this test ;).

olgabot: Yay, thank you!

ctb:
> My local `Cargo.lock` file has the differences below. I haven't committed them because they don't make sense to me. Do you know what may be going on?

it looks like maybe some cruft left over from using a development branch of sourmash, vs the official crates.io release. I would suggest this: try merging and then doing `cargo update -p sourmash` and see what happens!

ctb:
> I did that, but still get the same diff 😕 What I'm confused about is that `Cargo.lock` suggests using the official crate on `rust-lang`, but the current `main` of `sourmash_plugin_branchwater` matches my "old" one, what `Cargo.lock` is trying to change:

if you merge in main, and then change it to the rust-lang crate, you shouldn't get merge conflicts out of it?

In any case, I can deal with it in review, as long as your tests pass with whatever you have on the branch!

src/multisearch.rs
```diff
@@ -45,10 +168,10 @@ pub fn multisearch(
 
     let ksize = selection.ksize().unwrap() as f64;
 
-    let mut new_selection = selection;
+    let mut new_selection = selection.clone();
```
ctb, line 171: CTB: to check. This clone should maybe not be needed?

src/python/tests/test_fastgather.py
```diff
@@ -494,7 +494,11 @@ def test_against_multisigfile(runtmp, zip_against):
         "100000",
     )
     df = pandas.read_csv(g_output)
-    assert len(df) == 3
+    if zip_against:
```
ctb, line 497: I'm pretty sure this set of changes is unintentional. Can you revert to what's in main?

src/python/tests/test_sketch.py
```diff
@@ -1042,6 +1042,7 @@ def test_manysketch_prefix2(runtmp, capfd):
     for sig in sigs:
         assert sig.name in expected_signames
         if sig.name == "short":
+            # minhash is not defined? How does this test work? - @olgabot
```
ctb, line 1045: I think it only gets evaluated if the assert fails?

ctb: yep, verified! you can remove it if you like, or just leave it as is since it's not causing any harm.

src/multisearch.rs
ctb: I was correct that the clone is not required; I like not having it because it consumes `selection` so you can't reuse `selection` accidentally below.

ctb (changes requested):
Overall looks great! Only one or two minor changes left and then I can approve.

Note that I bumped sourmash to sourmash v0.17.1 in Cargo.toml.

ctb (approved):

src/multisearch.rs
olgabot: Yep, fixed! This saves a problem because I *was* using `selection` accidentally below .. so yay memory safety!

Issue #447, opened 2024-09-03
refactor tests to use `runtmp.sourmash(..., in_directory=...)` instead of try/finally


ctb:
in particular, test_fastmultigather.py has a bunch of these!

bluegenes:
https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/446 fixed these in `test_fastmultigather.py` only

PR #487, opened 2024-10-26
MRG: switch to using in_dir instead of try/finally for one last test

ctb:
Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/447


bluegenes (approved):
lgtm

Issue #398, opened 2024-07-25
fastmultigather returns matches only when --scaled is equal to scaled of the query


AnneliektH:
Hey, I am having some issues with fastmultigather and the scale of signatures:

Everything is in: `/group/ctbrowngrp2/scratch/annie/2023-sourmash-viruses`

I created protein sketches with the `custom_sketch.py`, that @ctb wrote. 
Sorry I cannot find your original repo, but [here](https://github.com/AnneliektH/2023-sourmash-viruses/blob/main/workflow/scripts/custom-sketch.py) is the script 

It is one sketch per all proteins for 1 organism, in this case a virus. 

I created sketches at 3 different ksizes (7,10,12), all at scaled 2. I then concatenated all 3 into one zip file: `/group/ctbrowngrp2/scratch/annie/2023-sourmash-viruses/results/signatures/240703_RefSeq.proteins.zip`

```
        python ../workflow/scripts/custom-sketch.py \
        {input.faa} --ksize {wildcards.ksize} --scaled 2 \
        -o {output.sig}
```

When I query this zipfile against the ICTV database using fastmultigather, I only get matches when the scaled is 2. I also want to use scaled=10 and scaled=100. I thought i could use lower scaled querys and that it automatically scales to the scale I am asking for. 

```
        sourmash scripts fastmultigather \
        ../results/signatures/240703_RefSeq.proteins.zip
         /home/ntpierce/2023-spillover/output.vmr/vmr_MSL38_v1.protein.zip \
        -k {wildcards.ksize} --scaled {wildcards.scaled} \
        -m protein -c {threads} -t 0 
```

Only works when --scaled 2. If 10 or 100, output is:
```
No matches to 'NC_020858.2'
No matches to 'NC_048199.1'
No matches to 'NC_003387.1'
```

For all signatures within the zip

I then created protein sketches with custom_sketch.py at scaled=10 and scaled=100, and used those as queries. If I do it that way, I do get matches from fastmultigather. 
Is this something that is not yet enabled for fastmultigather?

ctb:
ok, dug into this a little bit.

For a sketch against itself, with the original sketch at a scaled of 1000, you can't downsample like so:
```
sourmash scripts fastmultigather 47.sig.zip 47.sig.zip --scaled=10000
```
In addition, but slightly less? annoying, automatic downsampling doesn't work:
```
# create a 10k sig
[...]

PR #488, opened 2024-10-26
MRG: choose default scaled for multiple collections based on first collection

ctb:
This PR fixes an issue where `fastmultigather` and `multisearch` would fail to load sketches with scaled > 1000 if no `--scaled` parameter was set. It does so by making `scaled` an optional parameter to those commands; if `-s/--scaled` is not set, it is determined from the first collection loaded.

This PR also removes an unnecessary borrow of `selection` in all the various `do_` commands.

Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/398

TODO:
- [x] can `MultiCollection` determine scaled without the explicit loading of a sketch?
- [ ] is there a better way to handle `build_selection` if/else?
- [x] add tests
- [x] investigate multisearch, which might have same problem
- [ ] MAYBE switch all commands over to no default scaled value?

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -581,10 +581,6 @@ pub fn load_collection(
         Some((coll, n_failed)) => {
             let n_total = coll.len();
 
-            let foo = selection.scaled();
-            if let Some(scaled) = foo {
-                eprintln!("AAA: {}", scaled);
-            }
             let selected = coll.select(selection)?;
             let n_skipped = n_total - selected.len();
             report_on_collection_loading(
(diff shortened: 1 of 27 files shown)


src/utils/mod.rs
```diff
@@ -967,11 +972,18 @@ pub fn build_selection(ksize: u8, scaled: usize, moltype: &str) -> Selection {
     //     .map_err(|_| panic!("Unknown molecule type: {}", moltype))
     //     .unwrap();
 
-    Selection::builder()
-        .ksize(ksize.into())
-        .scaled(scaled as u32)
-        .moltype(hash_function)
-        .build()
+    if let Some(scaled) = scaled {
```
ctb, line 975: it seems like this could be done more cleanly?

ctb:
Ready for review & merge!

bluegenes (approved):
seems good to me

PR #493, opened 2024-11-04
EXP: support running flamegraph code

ctb:
add debug symbols, force specific python version for wheel building.


luizirber:
Have you tried `py-spy`? It also generates flamegraphs, and can resolve Python call stacks + native extensions properly:
https://github.com/benfred/py-spy?tab=readme-ov-file#record
https://github.com/benfred/py-spy?tab=readme-ov-file#can-py-spy-profile-native-extensions

Issue #491, opened 2024-10-28
check scaled bug on rocksdb-based fastmultigather


ctb:
does the fix in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/488 need to be applied to the rocksdb based fastmultigather?

PR #498, opened 2024-11-04
MRG: Fix scaled determination in `fastgather` and `fastmultigather`

ctb:
Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/491

This PR updates `fastgather` and `fastmultigather` against rocksdb to pick scaled automatically, rather than using a default value of 1000. `fastmultigather` now exits when no search sketches are found.

`scaled=1000` is removed as the default throughout.

It also eliminates the use of usize throughout the codebase in favor of u32/u64, following https://github.com/sourmash-bio/sourmash/pull/3364.

TODO:
- [x] test fail exit on fastmultigather with no matching foo
- [x] fix scaled selection for all the other commands, too
- [x] finish the usize job started in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/500


ctb:
Ready for review & merge @bluegenes !

src/fastgather.rs
```diff
@@ -12,8 +13,7 @@ use crate::utils::{
 pub fn fastgather(
     query_filepath: String,
     against_filepath: String,
-    threshold_bp: usize,
-    scaled: usize,
```
bluegenes, line 16: wait, so now it's not possible to pass in a different scaled -- the collection _must_ be at the scaled you want to search with?

bluegenes: oh nm, it's in selection, right?

bluegenes (approved):
lgtm

src/fastgather.rs
ctb: yep!

PR #499, opened 2024-11-04
WIP: debug/fix fastgather scaled problem

ctb:
Note: PR into https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/498

Puzzling mistake that points to an error in sourmash rust code somewhere. More work to be done.


ctb:
resolved! 😌 bug in my thinking.

PR #523, opened 2024-11-16
MRG: buildutils for singlesketch

bluegenes:
also use `buildutils` for singlesketch

To do:
- [x] actually report any errors coming back from sig building
- [x] add back sequence count reporting/sig reporting 
- [x] can we improve/abstract the sig writing more? e.g. pass filename in and handle the stdin vs file within the `BuildCollection`? 
- [x] Same for stdin vs filename -- handle internally
- [x] test gzipped JSON output
- [x] check gzipped output is actually gzipped
- [x] test zip output


src/singlesketch.rs
```diff
@@ -12,84 +8,34 @@ pub fn singlesketch(
...
-        // Write signatures to output file
-        let outpath = Path::new(&output);
-        let file = File::create(outpath)?;
-        let mut writer = BufWriter::new(file);
-
-        // Write in JSON format
-        serde_json::to_writer(&mut writer, &sigs)?;
-    }
+    let sequence_count =
+        sigs.build_sigs_from_file_or_stdin(&input_moltype, name, input_filename.clone())?;
```
mr-eyes, line 28: Given this new design, I will have to either modify the `build_sigs_from_file_or_stdin` or implement a new command. I will go for a new command to keep this `singlesketch` code clean. `parallelsinglesketch` is long, so I think `psinglesketch` or `parallelsketch`. cc @ctb

bluegenes: Hi Mo, feel free to modify the buildutils- it'd be great to have something reusable. 

I don't need to modify singlesketch like this, was just trying to get things to a modular and reusable state.

PR #531, opened 2024-11-20
MRG: add capacity for skipmer sketching and search

bluegenes:
Skipmers are something we've considered adding for quite some time, as DNA kmer that ~allows "mismatches" aka increases entropy + sensitivity.

Over in https://github.com/sourmash-bio/sourmash/pull/3395, I added a `skipm1n3` and `skipm2n3` moltypes, as well as code in `SeqToHashes` to build them. In https://github.com/sourmash-bio/sourmash/pull/3446/ I also added capacity for sourmash python function to read skipmer sigs, so `sig cat`, `sig summarize`, etc should now work.

There are two types of skipmers available, keep-2,skip-1 ("skipm2n3") and keep-1,skip-2 ("skipm1n3"). To sketch with skipmers, specify `skipm2n3` or `skipm1n3` in the parameter string. The skipmer ksize is the "final" size that the k-mer ends up. --i.e. for ksize 3, the sequence ACTAG would produce two skip-mers for m2n3: ACA, CTG.

example sketching commands:

manysketch:
```
sourmash scripts manysketch -p skipm2n3,k=21,scaled=1000 ms.csv -o output.zip
```

singlesketch:
```
sourmash scripts singlesketch -p skipm2n3,k=21,scaled=1000 myfile.fasta -o myfile.sig.gz
```
> `-o myfile.zip` also works

Skipmer References:
- [Skip-mers: increasing entropy and sensitivity to detect conserved genic regions with simple cyclic q-grams](https://www.biorxiv.org/content/10.1101/179960.abstract)
- [Extracting and Evaluating Features from RNA Virus Sequences to Predict Host Species Susceptibility Using Deep Learning](https://dl.acm.org/doi/abs/10.1145/3473258.3473271)

ref #549


bluegenes:
prior to merge, need to switch to sourmash 0.18.0 in Cargo.toml

ctb (approved):

Issue #538, opened 2024-12-04
sourmash scripts singlesketch output differs from sourmash sketch


peterjc:
I would expect ``sourmash scripts singlesketch`` and ``sourmash sketch`` to give identical signature files. However, they differ slightly (in the metadata).

Using viral genome https://github.com/pyani-plus/pyani-plus/blob/main/tests/fixtures/viral_example/OP073605.fasta as a test case. This is deliberately not in the current directory to clarify the issue (see below). Testing here on macOS with Python 3.12:

```console
❯ sourmash sketch dna -p 'k=31,scaled=300' viral_example/OP073605.fasta -o OP073605.sketch.sig 

== This is sourmash version 4.8.11. ==
== Please cite Irber et. al (2024), doi:10.21105/joss.06830. ==

computing signatures for files: viral_example/OP073605.fasta
Computing a total of 1 signature(s) for each input.
... reading sequences from viral_example/OP073605.fasta
calculated 1 signatures for 1 sequences in viral_example/OP073605.fasta
saved 1 signature(s) to 'OP073605.sketch.sig'. Note: signature license is CC0.
```

```console
❯ sourmash scripts singlesketch -I DNA -p 'k=31,scaled=300' viral_example/OP073605.fasta -o OP073605.singlesketch.sig 

== This is sourmash version 4.8.11. ==
== Please cite Irber et. al (2024), doi:10.21105/joss.06830. ==

=> sourmash_plugin_branchwater 0.9.11; cite Irber et al., doi: 10.1101/2022.11.02.514947

params: ['k=31,scaled=300,dna']
sketching file 'viral_example/OP073605.fasta' (DNA) with params 'k=31,scaled=300,dna' and name 'OP073605.fasta' using a single thread
calculated 1 signatures for 1 sequences in viral_example/OP073605.fasta
...singlesketch is done! results in 'OP073605.singlesketch.sig'
```

We expect the signatures to be identical, but they are not:

```console
❯ diff OP073605.sketch.sig OP073605.singlesketch.sig  
1c1
[...]

PR #548, opened 2024-12-18
MRG: update to (prospective) sourmash r0.18.0

ctb:
This PR updates to sourmash core latest, which will be r0.18.0.

The biggest set of changes in the plugin relate to https://github.com/sourmash-bio/sourmash/pull/3431, which made it possible to detect bad (and good) zip files based on content, and handle the error; in particular, this means that we can continue past errors from reading a file misnamed as a sig zip file, and try reading it in other ways. So a bunch of tests (e.g. `test_fastgather.py::test_bad_against_3`) no longer apply!

sourmash r0.18.0 also includes https://github.com/sourmash-bio/sourmash/pull/3434, which changes the behavior of `Signature::name()` to return an `Option` if `name` is not set, rather than defaulting to either `filename()` or `md5sum()`, in that order. This caused a bug (see https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/538#issuecomment-2543882346, split out into https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/550) where manifests would improperly disagree about whether two `Record`s matched. 

Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/550

TODO:
- [x] add test for manifest intersection per https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/550
- [x] should we add back a test for a misnamed zip file?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -206,7 +206,7 @@ dependencies = [
  "bitflags",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
+ "itertools 0.11.0",
  "lazy_static",
  "lazycell",
  "proc-macro2",
@@ -808,10 +808,11 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.72"
+version = "0.3.76"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a88f1bda2bd75b0452a14784937d796722fdebfe50df998aeb3f0b7603019a9"
+checksum = "6717b6b5b077764fb5966237269cb3c64edddde4b14ce42647430a78ced9e7b7"
 dependencies = [
+ "once_cell",
  "wasm-bindgen",
 ]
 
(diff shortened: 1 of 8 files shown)


ctb:
Ready for review & merge @bluegenes! But I will merge this w/o a review as needed in the interest of getting https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/531 into a release ;).

bluegenes (approved):
Lgtm

bluegenes:
We still have a test for improper zip file that can't be read, right?

ctb:
Uhh... actually, good question. Probably. What kind(s) of things are you thinking about? I could see -

* zip file that is corrupted
* zip file that doesn't have a manifest

other things?

bluegenes:
Yep -- I was mostly thinking about zipfiles that get interrupted while writing and thus can't be read - I guess this falls into your "corrupted" category. The error is now propagated from sourmash, so I'm guessing we just exit and say we can't load the sigs. 

Just wanted to make sure we still have a test covering this.

Issue #527, opened 2024-11-16
update code to use `min_max_scaled`


ctb:
from https://github.com/sourmash-bio/sourmash/issues/3386

PR #567, opened 2025-01-03
MRG: refactor fastgather/fastmultigather CSV output to use mpsc channels

ctb:
Note: PR into https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/568

This PR adds support for a single output file with `-o` for fastmultigather on a non-RocksDB database. It also refactors `fastgather` to use the same underlying mpsc mechanism, so all three gather output mechanisms share more code in common. The ultimate goal is to enable a better internal API, ref #569 and #552.

In particular, this means that now `-o` works the same way on a RocksDB database and on a non-RocksDB database :). 

This PR also disables the current functionality of creating individual output files for each query, which simplifies matters greatly, but _does_ break backwards compatibility. See long-ranging discussions over in https://github.com/sourmash-bio/sourmash/pull/2722 and https://github.com/sourmash-bio/sourmash/issues/2328.

Finally, one last breakage: `--create-empty-results` now _only_ creates empty prefetch results files, and not gather results files.

This PR also:
* Uses new sourmash `Collection::min_max_scaled` method (ref https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/527);
* makes sure `utils::prefetch(...)` only yields non-zero overlaps, preventing infinite loops;
* fixes clippy messages about complex return tuples/types;

This PR:
* Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/239
* Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/299
* Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/527

TODO:
- [x] adjust output_empty_results - ref https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/446
- [x] update docs
- [x] fix remaining tests
- [ ] write example/demo code to split results out by query

### DIFF
--- a/src/python/sourmash_plugin_branchwater/__init__.py
+++ b/src/python/sourmash_plugin_branchwater/__init__.py
@@ -278,7 +278,6 @@ class Branchwater_Fastmultigather(CommandLinePlugin):
             "-o",
             "--output",
             help="CSV output file containing gather matches",
-            required=True,
         )
         p.add_argument(
             "--create-empty-results",


ctb:
Ready for review @bluegenes!

doc/README.md
```diff
@@ -327,32 +327,28 @@ this can be a significant time savings for large databases.
...
 #### Output files for `fastmultigather`
 
-On a database of sketches (but not on RocksDB indexes)
-`fastmultigather` will output two CSV files for each query, a
-`prefetch` file containing all overlapping matches between that query
-and the database, and a `gather` file containing the minimum
-metagenome cover for that query in the database.
+`fastmultigather` will output a single file containing all gather
+results, specified with `-o/--output`. No prefetch results will be
+output.
```
bluegenes, line 332: this is a little confusing with the prefetch output note below

src/python/sourmash_plugin_branchwater/__init__.py
```diff
@@ -277,13 +277,14 @@ def __init__(self, p):
         p.add_argument(
             "-o",
             "--output",
-            help="CSV output file for matches. Used for non-rocksdb searches only.",
+            help="CSV output file for matches.",
         )
         p.add_argument(
             "--create-empty-results",
```
bluegenes, line 283: My only note here is that I would now (given single output file) interpret this as writing empty lines in the output fastmultigather results file, which we don't do...

bluegenes (approved):
a couple documentation qualms - otherwise lgtm 🎉

Issue #565, opened 2025-01-03
`test_sketch.py::test_manysketch_skipm2n3` is failing on main?


ctb:
`src/python/tests/test_sketch.py::test_manysketch_skipm2n3`

```
>       assert len(sigs) == 3  # 3 dna, 3 skipmer. But sourmash can only read the DNA sigs!!
E       AssertionError: assert 6 == 3
E        +  where 6 = len([SourmashSignature('short3', f85747ac), SourmashSignature('short3', 890557b3), SourmashSignature('short', 1474578c), SourmashSignature('short', 0486fcae), SourmashSignature('short2', 4efeebd2), SourmashSignature('short2', ec6305f5)])
```

not sure if it's my laptop, or what.

bluegenes:
Ah, this test assumed we could not read skipmer signatures with sourmash python layer. Needs to be updated now that we ~can.

ctb:
Fixed in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/568

PR #568, opened 2025-01-04
MRG: Fix skipmer test `test_manysketch_skipm2n3` now that sourmash is updated

ctb:
Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/565

### DIFF
--- a/src/python/tests/test_sketch.py
+++ b/src/python/tests/test_sketch.py
@@ -1471,12 +1471,17 @@ def test_manysketch_skipm2n3(runtmp, capfd):
     sigs = list(idx.signatures())
     print(sigs)
 
-    assert len(sigs) == 3  # 3 dna, 3 skipmer. But sourmash can only read the DNA sigs!!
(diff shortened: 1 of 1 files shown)


bluegenes (approved):
Lgtm!

ctb:
Wait... are the tests failing on the CI because they're using the released version of sourmash, which doesn't fully support skip-mers? So we need to release a new version of sourmash to get this to work? 😭

bluegenes:
sigh, i think so. Also it needs https://github.com/sourmash-bio/sourmash/pull/3457. The fixes I have there fix `sig downsample`, but I'm not sure they fix every issue, since I don't have comprehensive skipmer testing on the sig functions

ctb:
k - for now, I'm going to "fix" this by reverting to the old behavior, but with a comment. Then when it starts breaking again on the new sourmash release, will fix it again. Frustrating situation but no simple way around it 😅

PR #569, opened 2025-01-05
MRG: refactor the internal Rust API to support container-level functionality

ctb:
This PR splits out an internal Rust API into new `_obj` functions for many of the search commands in the branchwater plugin, to provide better separation between the CLI and the internal Rust API. The ultimate goal is to build out an intermediate container-level Python API in #552, but that was getting messy between the Rust changes, pyo3 layer additions, and Python tests! Hence, this PR, which will probably involve minimal changes to the CLI and the tests. See #552 for some of the cool new stuff this enables.

This is all refactoring, no new functionality.

A few other features of note - this PR
* sequesters most of the `println` and `eprintln` progress & error reporting to the CLI-level stuff
* switches rare/unexpected errors over to `expect` instead of raising an explicit `Err`
* removes all of the `Box<dyn std::error::Error>` stuff, and manages it via a mixture of `anyhow` and `?` conversion
* changes the container-level layer to report useful numbers (`n_processed` etc) while keeping the CLI layer focused on reporting exit codes

TODO:
- [x] remove out all the `@CTB` notes
- [x] sequester the new `clone`s out


bluegenes (approved):
Nice. The additional layer of abstraction seems pretty clean here, excited for what it can enable 🎉

ctb:
cleanliness FTW! we've learned so much 😆

PR #573, opened 2025-01-06
MRG: add tests for `stdin` input to `singlesketch` 

### LINKED ISSUE #572: add `stdin` tests for `singlesketch`
No tests cover `stdin` input for `singlesketch`

ref #571

bluegenes:
- Use needletail `parse_fastx_stdin` instead of `parse_fastx_reader`
- Add test for `stdin` to `singlesketch`
  - fixes #572  
- Fix `singlesketch` tests that were checking 0 hashes (using scaled=1000 does not work for most small test fasta)

### DIFF
--- a/src/python/tests/test_sketch.py
+++ b/src/python/tests/test_sketch.py
@@ -3,6 +3,7 @@ import pytest
 import csv
 import pandas
 import sourmash
+import subprocess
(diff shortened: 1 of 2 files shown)


bluegenes:
ready for review @ctb

PR #577, opened 2025-01-08
MRG: test generated manifest 

bluegenes:
It turns out the manifest test were re-building the manifest from the sigs, and not testing the manifest in the zip file. This PR rectifies that and fixes the code for `n_hashes`.


bluegenes:
@ctb ready for review

ctb (approved):

Issue #580, opened 2025-01-09
"Error: Invalid argument: Column family not found: storage" when searching old (v0.9.5) index with latest (v0.9.12) plugin


matthuska:
I have an index that was created using sourmash_plugin_branchwater v0.9.5 which I could search without any errors. After updating the plugin to v0.9.12, I now get the error:

```
Error: Invalid argument: Column family not found: storage
```

The search command being used is:

```
$ sourmash scripts manysearch --ksize 21 --moltype DNA --scaled 1000 --cores 1 --threshold 0.9 --output <whatever> input.sig path/to/old/rocks.db/
```

@luizirber already found the change that is causing the error (roughly: a new rockdb column was added, which is supposed to automatically be added to older indexes but the index is being opened read-only).

ctb:
replicated locally!

PR #416, opened 2024-08-12
MRG: update documentation for RocksDB `index` and internal/external storage, + miscellaneous improvements

ctb:
Tackling #409 and documenting RocksDB `index` internal/external storage per #390.

Items from #409:
- [x] internal/external storage fuss and muss
- [x] add `index` command to TOC
- [x] add `multisearch` threshold documentation
- [x] add pretty-printing info from #408
- [x] update `manysketch` fromfile formatting discussion
- [x] update `manysketch` multithreading discussion

This PR also updates the docs to:
- use 'pathlist' instead of 'fromfile' per discussion in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/354#issuecomment-2200095830 - syncing with sourmash, viz https://github.com/sourmash-bio/sourmash/pull/3027
- properly discuss `manysearch` column presence/absence as adjusted in https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/408

Fixes #409.

---

Here are some of the bigger/more confusing changes, for reviewers to evaluate and help improve ;) --

## multisearch threshold discussion:

>The `-t/--threshold` for `multisearch` and `pairwise` applies to the
containment of query-in-target and defaults to 0.01.  To report
_any_ overlap between two sketches, set the threshold to 0.

## manysearch output discussion

> The results file here, `query.x.gtdb-reps.csv`, will have the following columns: `query`, `query_md5`, `match_name`, `match_md5`, `containment`, `jaccard`, `max_containment`, `intersect_hashes`, `query_containment_ani`.

>If you run `manysearch` _without_ using a RocksDB database (that is, against regular sketches), the results file will also have the following columns: , `match_containment_ani`, `average_containment_ani`, and `max_containment_ani`.

>Finally, if using sketches that have abundance information, the
results file will also contain the following columns: `average_abund`,
`median_abund`, `std_abund`, `n_weighted_found`, and `total_weighted_hashes`.

>See
[the prefetch CSV output column documentation](https://sourmash.readthedocs.io/\
en/latest/classifying-signatures.html#appendix-e-prefetch-csv-output-columns)
for information on these various columns.

## Internal vs external storage of sketches in a RocksDB index

(The below applies to v0.9.7 and later of the plugin; for v0.9.6 and
before, only external storage was implemented.)

RocksDB indexes support containment queries (a la the
[branchwater application](https://github.com/sourmash-bio/branchwater)),
as well as `gather`-style mixture decomposition (see
[...]

PR #581, opened 2025-01-09
MRG: support `check --upgrade` to upgrade old versions of RocksDB/RevIndex

ctb:
This PR deals with the ramifications of a format change for RocksDB/RevIndex databases that probably occurred around v0.9.7, in which the 'storage' column was added to RevIndex databases. Prior to that version, RevIndex databases always stored sketches externally; see https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/390, https://github.com/sourmash-bio/sourmash_plugin_branchwater/pull/416, and https://github.com/sourmash-bio/sourmash/pull/3250.

The addition of the 'storage' column breaks when trying to open old RocksDB/RevIndex databases with more recent versions of the plugin; the error message is: `Invalid argument: Column family not found: storage`.

This PR:
* updates `manysearch` and `fastmultigather` so that a slightly better error message is reported;
* upgrades `index` with an option `--upgrade` that will open when searching against a RocksDB, they will automatically upgrade the internal RocksDB format from older versions. This is as simple as opening them in read/write mode instead of read-only mode, which permits sourmash to upgrade the RevIndex;
* adds tests for RocksDB/RevIndex databases created with v0.9.5 and v0.9.13 (with both internal and external storage) so that plugin tests will properly detect database format changes.

An alternative solution would be to always open the database in rw mode so that it just gets upgraded automatically, but  it gives me the heeby jeebies to open a database in rw mode unnecessarily...

Fixes https://github.com/sourmash-bio/sourmash_plugin_branchwater/issues/580

TODO
- [ ] create issue with error message as doc bait


bluegenes:
I like the idea of having `check` update them! Can we output a warning from standard `check` and then have a flag, `--update` to go ahead and update?

ctb:
> I like the idea of having `check` update them! Can we output a warning from standard `check` and then have a flag, `--update` to go ahead and update?

implemented!

bluegenes (approved):
Lgtm!

PR #601, opened 2025-02-08
Documentation update: `manysketch --force` allows use of FASTA files in more than one sketch

mlovci:
Updated documentation to reference the behavior of `--force` in do_manysketch to indicate this option allows "use of individual FASTA files in more than  _one_ sketch"


olgabot (approved):
Thanks for catching and fixing!

ctb (approved):

ctb:
thank you!

PR #635, opened 2025-04-02
MRG: update zip version to 2.5

### LINKED ISSUE #389: Check in to see if we can update zip crate
When the bug found during #385 is fixed in zip, we should update zip and take it off the dependabot ignore list

bluegenes:
zip 2.3.1 had the bug, so they must have meant it was fixed in v2.5. Seems to work now for me!

- fixes #389

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2383,16 +2383,17 @@ dependencies = [
 
 [[package]]
 name = "zip"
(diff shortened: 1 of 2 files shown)


ctb:
awesome! go for it!

ctb (approved):

PR #641, opened 2025-04-22
MRG: report % of target matched by query

bluegenes:
For a project, we'd like to know the fraction of each metagenome matched by the query. 

I propose adding two columns to `manysearch` results:
- `containment_target_in_query` -  the % of the unique metagenome content matching the query
- `f_weighted_target_in_query` - the % of weighted metagenome content matching the query

We were already calculating the first, just not reporting it. Getting the second (`f_weighted_target_in_query`) is straightforward if we're already calculating the abundance stats, I think: `abundance-weighted sum of intersected hashes` / `abundance-weighted sum of all metagenome hashes`. 

`manysearch` doesn't have a single number for the reverse abundance stat: instead we have `avg_abund`, `med_abund`, `std_abund` for the query. But that's because we're dividing by the number of intersected hashes to get an abundance estimate for the query, rather than finding the total % of all metagenome hashes, as in this case.

### DIFF
--- a/src/manysearch.rs
+++ b/src/manysearch.rs
@@ -26,6 +26,7 @@ type AbundanceStats = (
     Option<f64>,
     Option<f64>,
     Option<f64>,
+    Option<f64>,
 );
 
 pub fn manysearch(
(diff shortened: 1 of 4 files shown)


bluegenes:
@ctb I think this one may actually be done already? ready for review!

ctb (approved):

PR #646, opened 2025-05-01
WIP: add average_containment and cosine sim to pairwise and multisearch

bluegenes:
I'm actually pretty surprised we left out average containment - was there a reason for this? 

Here, I've added `average_containment` as a standard output, and `cosine` as an optional output to be added if `--calc-abund-stats` is used and both query and against have abundances.

To do:
- dont check against abund on every comparison
- add tests for avg containment, cosine outputs
- add tests for using these for cluster


ctb:
> I'm also thinking it might be useful to add abundance-projected containment?

we'd have to add it in both directions, though - it's not a symmetric measure.

PR #654, opened 2025-06-09
upd rust to 1.75 for zip compatibility

bluegenes:
- update rust to 1.75 in `Cargo.toml`
- also pin rust to  1.75 in `environment.yml` to make sure we're installing the pinned rust version for tests.

Actually, we were already installing 1.75 here! So we were on 1.75 for tests already: https://github.com/sourmash-bio/sourmash_plugin_branchwater/blob/main/.github/workflows/build-test.yml#L47C11-L47C38

This puts us out of sync with base sourmash (1.74), but it doesn't seem to present an issue here.

@ctb thoughts on pinning version in `environment.yml` vs in the `build-test.yml` workflow (or keeping both)?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2373,13 +2373,12 @@ dependencies = [
 
 [[package]]
 name = "zip"
-version = "2.5.0"
+version = "4.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


bluegenes:
@ctb ready for review

ctb (approved):

ctb:
> @ctb thoughts on pinning version in `environment.yml` vs in the `build-test.yml` workflow (or keeping both)?

No clear idea :)

